PR #2, opened 2019-07-23
Fix typos and change markdown to be more accurate and readable.

daron666:


README.md
```diff
@@ -4,42 +4,28 @@
...
+```
+
+Of course you can also specify the exact subproject that you wanna add to yours dependencies (used in place of `"tofu"`):
+
+* `tofu-core` for core (Main core)
+* `tofu-memo` for memo (Caching utils)
+* `tofu-env` for env (Some variation of a Reader Monad)
+* `tofu-parallel` for parallel (Parallel utils)
+* `tofu-concurrent` for concurrent (Concurrent utils)
+* `tofu-optics"0.1"` for optics core (Optics typeclasses)
```
eld0727: I think that you add some new typos :)

daron666: sticker_r6nsqnag

daron666: Ok, will fix it. Like how much time one needs to fix all typos? Infinity.

PR #3, opened 2019-07-23
Fix typos that have been introduced by fixing typos.

daron666:
You wouldn't believe that. But I think I mastered that readme to be perfect.


daron666:
On the other hand that means here will be 3 commits from me. Like a real contributor.
jake

PR #5, opened 2019-08-07
optics-macro module

danslapman:
This module is actually a `monocle-macro` port

PR #7, opened 2019-08-09
Fix missing dependency

catostrophe:
The build depends on an unpublished version of `derevo`. Change version to the latest one available on Central.


danslapman:
Also fixed in #5

PR #11, opened 2019-08-28
Deprecation of 2.11, parallel and addition of optics syntax

Odomontois:


Z1kkurat:
Since this PR has some major changes (bumping cats version, removing deprecated typeclasses), it would be nice to rename it so the name could reflect PR's content (or at least add some description to it).

env/src/main/scala/tofu/env/Env.scala
```diff
@@ -184,7 +184,9 @@ sealed trait Env[E, +A] {
   def executeOn(s: Scheduler, forceAsync: Boolean = true): Env[E, A] =
     mapTask(_.executeOn(s, forceAsync))
   def fork: Env[E, Fiber[Env[E, *], A @uv]] = executeAsync.start
-  def forkAndForget: Env[E, Unit]           = mapTask(_.forkAndForget)
+  @scala.deprecated("Replaced with startAndForget", "0.2.1")
```
Z1kkurat: Shouldn't it be '0.3.0'?

env/src/main/scala/tofu/env/EnvTraversing.scala
```diff
@@ -1,66 +1,64 @@
 package tofu.env
 
 import monix.eval.Task
-
-import scala.collection.TraversableLike
-import scala.collection.generic.CanBuildFrom
+import monix.execution.compat.BuildFrom
```
danslapman: I think we should use official `scala.collection.compat.BuildFrom` instead

Odomontois: I can't see how we might use it having monix API is built upon that

Odomontois: I'll try

Odomontois: No. Scala 2.12.9 wont build

danslapman: Okay, sad though

PR #12, opened 2019-09-06
Config

Odomontois:


core/src/main/scala/tofu/syntax/monadic.scala
```diff
@@ -0,0 +1,49 @@
+package tofu.syntax
+import cats.syntax._
+import cats.{Apply, FlatMap, Functor}
+
+object monadic extends TupleSemigroupalSyntax with ApplicativeSyntax with MonadSyntax {
+  implicit class TofuFunctorOps[F[_], A](private val fa: F[A]) extends AnyVal {
```
danslapman: Just wonder, why do we need that? Isn't `cats.syntax.*` enough?

Odomontois: There is couple of reasons for such syntax.
1. cats uses simulacrum.typeclass macro for most operations, javap approved that it's generated syntax extension leads to allocation for each operation involved, that is : unneeded allocation for each line in your for-comprehension
2. as specific syntax import lover, I hate to write 
```
import syntax.functor._ 
import syntax.apply._
import syntax.applicative._ 
import syntax.flatMap._
import syntax.monad._
```
to have just simple support for monads

danslapman: ✅

PR #15, opened 2019-10-01
Make >> operator by-name like Cats's version is

Z1kkurat:


env/src/main/scala/tofu/env/Env.scala
```diff
@@ -70,7 +70,7 @@ sealed trait Env[E, +A] {
     mapTask3(eb, ec)(Task.parMap3(_, _, _)(f))
 
   def zip[B](t: Env[E, B]): Env[E, (A, B)]                 = map2(t)((a, b) => (a, b))
-  def >>[B](t: Env[E, B]): Env[E, B]                       = map2(t)((_, x) => x)
+  def >>[B](t: => Env[E, B]): Env[E, B]                    = flatMap(_ => t)
```
eld0727: What is motivation of this change?
Env is referential transparent and could be passed by value just like Task

Z1kkurat: The motivation for this PR is making the behaviour the same as Cats version. Using this operator, one could hope that it behaves exactly like Cats version, but it doesn't, so this may cause unexpected behaviour in case when used with recursive creation of argument

Z1kkurat: Consider the following code:
```
def go: Env[Unit, Unit] = Env.unit[Unit] >> go
go.run(())
```
This code will fail with StackOverflowError, thus making it impossible to define recursive calls.
Meanwhile this code runs just fine and produces value of Env type:
```
import cats.Monad
import cats.syntax.flatMap._
import tofu.env.Env

def go[F[_]: Monad]: F[Unit] = Monad[F].unit >> go
go[Env[Unit, Unit]].run(())
```

Z1kkurat:
I added an eager version of the same operation(`*>`) to support both evaluation types

env/src/main/scala/tofu/env/Env.scala
```diff
@@ -70,7 +70,8 @@ sealed trait Env[E, +A] {
...
+  def *>[B](t: Env[E, B]): Env[E, B]                       = map2(t)((_, x) => x)
```
Odomontois: What about `<<` and `<*` and possibly `&>` `<&` for `parMap2` ?

Z1kkurat: `<<` is already there and it seems there is no way to avoid eager evaluation of second argument since it must be evaluated first.
I will add the operators you mentioned and maybe a few more that can be handy (any ideas for those?)

Odomontois: `<<` can be made lazy on pseudo-this parameter via extension like
```scala 
implicit class LazyEnvOps[E, A](private self: => Env[E, A])
```

Z1kkurat: Can you please elaborate on that?
I tried something like this and it didn't work:
```scala
implicit class LazyEnvOps[E, A](self: => Env[E, A]) {
  def <<[B](that: Env[E, B]): Env[E, A] = that.flatMap(_ => self)
}

def env1: Env[Unit, Unit] = {
  println("env1 evaluated")
  Env.unit[Unit]
}

def env2: Env[Unit, Unit] = {
  println("env2 evaluated")
  Env.unit[Unit]
}

env1 << env2
```
It prints both statements. Am I missing something?

Odomontois: Well you may do something like
```scala
implicit class LazyEnvOps[E, A](self: => Env[E, A]) {
  def <<[B](that: Env[E, B]): Env[E, A] =  Env.defer(self) <* that
}
```

On the other hand, it would be more rational to make it lazy on the second argument 

```scala

def flatTap[B](that: A => Env[E, B]): Env[E, A] = flatMap(x => that as x)
def <<[B](that: => Env[E, B]): Env[E, A] = flatTap(_ => that)
```

Z1kkurat: Sorry I misunderstood designation of `<<` at first.
I added `flatTap` and operators for `parMap2`. Any ideas for more operators?

PR #21, opened 2019-10-25
Add basic Env docs and example

Z1kkurat:


Odomontois:
I think it would be more useful in https://github.com/TinkoffCreditSystems/tofu/tree/master/docs  subproject.

PR #31, opened 2019-11-11
Add string interpolators for error logging

ioaoue:
As of SLF4J version 1.6.0, it is possible to log an exception along with a parameterized message. This PR adds some interpolators to use that feature.

### DIFF
--- a/logging/structured/src/main/scala/tofu/logging/Logging.scala
+++ b/logging/structured/src/main/scala/tofu/logging/Logging.scala
@@ -33,8 +33,8 @@ trait LoggingBase[F[_]] {
     write(level, message, values: _*)
 
   /** could be overridden in the implementations, write message about some exception */
-  def writeCause(level: Level, message: String, cause: Throwable): F[Unit] =
-    write(level, message, Logging.errorvalue(cause))
+  @silent def writeCause(level: Level, message: String, cause: Throwable, values: LoggedValue*): F[Unit] =
+    write(level, message, values: _*)
 
   def trace(message: String, values: LoggedValue*): F[Unit] = write(Trace, message, values: _*)
   def debug(message: String, values: LoggedValue*): F[Unit] = write(Debug, message, values: _*)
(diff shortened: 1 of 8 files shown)


logging/structured/src/main/scala/tofu/logging/impl/ContextLoggingImpl.scala
```diff
@@ -33,14 +33,14 @@ class ContextLoggingImpl[F[_]: Applicative, C: Loggable, Service: ClassTag](cont
...
+  override def traceCause(message: String, cause: Throwable, values: LoggedValue*): F[Unit] =
+    context.ask(ctx => logger.trace(ContextMarker(ctx), message, values :+ cause: _*)).whenA(errorEnabled)
+  override def debugCause(message: String, cause: Throwable, values: LoggedValue*): F[Unit] =
+    context.ask(ctx => logger.debug(ContextMarker(ctx), message, values :+ cause: _*)).whenA(errorEnabled)
+  override def infoCause(message: String, cause: Throwable, values: LoggedValue*): F[Unit] =
+    context.ask(ctx => logger.info(ContextMarker(ctx), message, values :+ cause: _*)).whenA(errorEnabled)
+  override def warnCause(message: String, cause: Throwable, values: LoggedValue*): F[Unit] =
+    context.ask(ctx => logger.error(ContextMarker(ctx), message, values :+ cause: _*)).whenA(errorEnabled)
+  override def errorCause(message: String, cause: Throwable, values: LoggedValue*): F[Unit] =
+    context.ask(ctx => logger.error(ContextMarker(ctx), message, values :+ cause: _*)).whenA(errorEnabled)
```
ioaoue: I don't know how this change will affect performance

logging/structured/src/main/scala/tofu/logging/Logging.scala
```diff
@@ -33,8 +33,8 @@ trait LoggingBase[F[_]] {
     write(level, message, values: _*)
 
   /** could be overridden in the implementations, write message about some exception */
-  def writeCause(level: Level, message: String, cause: Throwable): F[Unit] =
-    write(level, message, Logging.errorvalue(cause))
+  @silent def writeCause(level: Level, message: String, cause: Throwable, values: LoggedValue*): F[Unit] =
+    write(level, message, values: _*)
```
ioaoue: I think the previous version could do something strange. So I changed it a bit.

logging/structured/src/main/scala/tofu/logging/impl/ContextLoggingImpl.scala
Odomontois: I don't understand the reason why we need to append `cause` to values here. It seems counter-intuitive

Odomontois: We'll have the `cause` in the log message available to a layout anyway.

logging/structured/src/main/scala/tofu/logging/Logging.scala
Odomontois: Now we lost the `cause`, I suppose appending `values :+ Logging.errorvalue(cause)` would be more suitable here

logging/structured/src/main/scala/tofu/logging/impl/ContextLoggingImpl.scala
Odomontois: Oh. I get it. logger doesn't have API for messages with interpolation as well as cause.
I suppose we must not throw cause argument here and break compatibility with existing formats handling causes.
Instead, you can do some wrapping values in the Marker

logging/structured/src/main/scala/tofu/logging/Logging.scala
Odomontois: I propose next changes:

1. Move `Logging.errorvalue` to `LoggedValue.error`
2. Change the result of `LoggedValue.error` to the new wrapper
   that is sublclass of both `LoggedValue` and `Throwable(cause.message, cause)`.

PR #32, opened 2019-11-13
console typeclass

Odomontois:


danslapman:
Not sure, if `common` is a good place for it. Maybe plase it in `env` module?

Odomontois:
> Not sure, if `common` is a good place for it. Maybe plaсe it in `env` module? (I suppose it designed to be used as a context)

`env` is a specific package for `Env` data type
I suppose `console` operations are pretty `common`. 
I'd like all the basic Sync.delay cases to be covered by `tofu.core` and to have  implemetation for the upcoming `Sim` data type.
Ultimate goal here to get rid of every possible source of indeterminism in tests.

Issue #28, opened 2019-11-10
LogginImpl stackoverflow


optician:
```
import cats.effect.IO
import org.slf4j.LoggerFactory
import tofu.logging.Logging
import tofu.logging.impl.LoggingImpl

object HerdingAnimals {
  def main(args: Array[String]): Unit = {
    implicit val logger: Logging[IO] = new LoggingImpl[IO](LoggerFactory.getLogger(getClass.getName)) {}
    logger.info("test").unsafeRunSync()
  }
}
```
Runtime:
```
Exception in thread "main" java.lang.StackOverflowError
	at tofu.logging.impl.LoggingImpl.write(LoggingImpl.scala:17)
	at tofu.logging.LoggingBase.info(Logging.scala:41)
	at tofu.logging.LoggingBase.info$(Logging.scala:41)
	at tofu.logging.impl.LoggingImpl.info(LoggingImpl.scala:8)
	at tofu.logging.impl.LoggingImpl.write(LoggingImpl.scala:20)
	at tofu.logging.LoggingBase.info(Logging.scala:41)
	at tofu.logging.LoggingBase.info$(Logging.scala:41)
	at tofu.logging.impl.LoggingImpl.info(LoggingImpl.scala:8)
	at tofu.logging.impl.LoggingImpl.write(LoggingImpl.scala:20)
	at tofu.logging.LoggingBase.info(Logging.scala:41)
	at tofu.logging.LoggingBase.info$(Logging.scala:41)
```

Odomontois:
LoggingImpl is not supposed to used that way.
I assume we need to make it package private

PR #38, opened 2019-11-20
refactor elk logging

Odomontois:
make LoggingImpl package private


Odomontois:
From now on ELKLogger can be configured via 
```xml
<layout class="tofu.logging.ELKLayout">
  <argumetsField>args</arguments>
 </layout>
```
in logback.xml

That will activate embedding arguments as different objects inside the `args` array

Odomontois:
Also fixes #28

PR #39, opened 2019-11-21
Grouping logging values

Odomontois:
Also recalculating loggable instance


Odomontois:
@Z1kkurat How are you doing, lad?

PR #43, opened 2019-11-27
Add mode to @masked annotation

REDNBLACK:
1. `MaskMode.Erased` acts like `erased == true`
2. `MaskMode.Full` acts like `erased == false`
3. `MaskMode.ForLength(offset, maxLength)`
starts masking from `offset` until `maxLength` count of symbols were masked. `maxLength == -1` acts like total string length,

### DIFF
--- a/logging/derivation/src/main/scala/tofu/logging/derivation/annotations.scala
+++ b/logging/derivation/src/main/scala/tofu/logging/derivation/annotations.scala
@@ -6,7 +6,8 @@ import scala.annotation.StaticAnnotation
   * when applied to case class means "logShow should be empty string" */
 final case class hidden() extends StaticAnnotation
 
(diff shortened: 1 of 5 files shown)


Odomontois:
Cool extension!
I wonder if we can add regex mode too, with the meaning that the first matched group if any should be masked. 
So `ForLength(n)` would become `Regex(".{n}(.*)")` and `ForLength(n, m)` would become `Regex(".{n}(.{1,m}).*)`

Odomontois:
@REDNBLACK Are you interested to continue to work on this?

REDNBLACK:
Yeah

Odomontois:
If it's possible can you look into this issue #59 as wel

Odomontois:
fixed 59

PR #44, opened 2019-11-27
Setup inlining, mark all implicit value classes as final with private val

REDNBLACK:


build.sbt
```diff
@@ -34,7 +34,7 @@ lazy val defaultSettings = List(
   libraryDependencies += compilerPlugin("com.olegpy"    %% "better-monadic-for" % "0.3.1"),
   libraryDependencies += scalatest,
   libraryDependencies ++=
-    List(
+    Seq(
```
Odomontois: What is accomplised here? Actually I can't see significant difference

concurrent/src/main/scala/tofu/concurrent/Daemon.scala
```diff
@@ -88,11 +88,11 @@ object Daemon extends DaemonInstances {
     }
   }
 
-  implicit class DaemonOps[F[_], E, A](val self: Daemon[F, E, A]) extends AnyVal {
+  implicit final class DaemonOps[F[_], E, A](private val self: Daemon[F, E, A]) extends AnyVal {
```
Odomontois: What it actually does. I've thought whole `AnyVal` buisness here is not allocate any Ops, so what difference `final` modifier does?

env/src/main/scala/tofu/env/EnvSpecializedFunctions.scala
```diff
@@ -179,7 +179,7 @@ trait EnvSpecializedFunctions[E] {
   }
 
   object converters {
-    implicit class TaskAsEnv[A](val task: Task[A]) {
+    implicit class TaskAsEnv[A](private val task: Task[A]) {
```
Odomontois: That's good catch

build.sbt
REDNBLACK: I remember that author of sbt recommended usage of Seq, but i did it just for consistency ¯\_(ツ)_/¯

concurrent/src/main/scala/tofu/concurrent/Daemon.scala
REDNBLACK: final is for the better chance of inlining

REDNBLACK:
@Odomontis One thing, do we care about bincompat, or planning to setup sbt-mima plugin for future releases of tofu? If yes, then we shouldn’t enable “opt:l:inline” (cross-method inlining).

concurrent/src/main/scala/tofu/concurrent/Daemon.scala
Odomontois: But those operations are never exposed as methods of classes, they should be inlined by the compiler itself

PR #84, opened 2019-12-20
WIP Bifunctor Env alternative

Z1kkurat:
This is very early WIP PR, almost nothing is implemented yet.
If you have any feedback concerning design or implementation (esp. regarding performance) - I would be very glad to hear and try to address it as early as possible.


env/src/main/scala/tofu/env/bio/EnvBio.scala
```diff
@@ -0,0 +1,42 @@
+package tofu.env.bio
+
+import monix.eval.Task
+
+sealed trait EnvBio[-R, +E, +A] {
+  def run(ctx: R): Task[Either[E, A]]
```
Odomontois: I'm not sure if we need `sealed` here
Constructing such element using just lambda to `Task[Either[E, A]]` should be nice

env/src/main/scala/tofu/env/bio/EnvBio.scala
```diff
@@ -0,0 +1,42 @@
...
+  def run(ctx: R): Task[Either[E, A]] = Task.defer(runF(ctx)).attempt.flatMap {
+    case Left(UserError(err: E)) => Task.pure(Left(err))
+    case Left(fatalErr)          => Task.raiseError(fatalErr)
+    case Right(Left(userErr))    => Task.pure(Left(userErr))
+    case Right(Right(value))     => Task.pure(Right(value))
+  }
+}
+
+/** Context aware variation of `Env` that has no notion of error type, thus all errors to underlying 'Task' are considered fatal */
+final case class EnvUioCtx[R, +A](runF: R => Task[A]) extends EnvBio[R, Nothing, A] {
```
Odomontois: I would greatly recomment not wrapping functions inside case classes
It would just have better ergonomics and performance if `runF` were just an abstract method

env/src/main/scala/tofu/env/bio/EnvBio.scala
```diff
@@ -0,0 +1,42 @@
...
+  def run(ctx: R): Task[Either[Nothing, A]] = Task.defer(runF(ctx).map(Right.apply))
```
Odomontois: If `runF` is just a method `defer` is redundant here

env/src/main/scala/tofu/env/bio/EnvBio.scala
```diff
@@ -0,0 +1,42 @@
...
+
+  private def raiseUserError[E1](e: E1): Task[Either[E1, Nothing]] = Task.raiseError(UserError(e))
+}
+
+object EnvBio extends EnvBioFunctions {}
+
+/** Context aware variation of `EnvBio` that allows specifying error type */
+final case class EnvBioCtx[R, +E, +A](runF: R => Task[Either[E, A]]) extends EnvBio[R, E, A] {
+  def run(ctx: R): Task[Either[E, A]] = Task.defer(runF(ctx)).attempt.flatMap {
+    case Left(UserError(err: E)) => Task.pure(Left(err))
```
Odomontois: With this is encoding is this even possible that `runF` will ever raise `UserError` ?

Z1kkurat: My initial idea was that computations can be short-circuited by `EnvBio.raiseError` that raises `UserError`. But now, I realise that `run` is called not only at the end of program, but can be called somewhere in the middle of computation, so that wasn't the best idea and it even prevents us from mapping on error channel. I will rework this to just passing Either around.

Odomontois: I dont think people will like wrapping Eithers on every step

Z1kkurat: I agree. Then internal encoding can be represented as `A => Task[B]` and user custom errors as `UserError <: Throwable`, so in error handling we will match error and handle user ones. Thanks!

Z1kkurat:
@Odomontois  I reimplemented underlying representation considering your advices

Z1kkurat:
Current status: all basic functionality is supported - delaying effects, handling errors (including working with error channel), access to context and some other minor features. I don't know if this can be considered out of WIP, probably yes, and I will port other functions from the main Env implementation afterwards.

PR #85, opened 2019-12-20
Make ServiceLogging invariant

Odomontois:


logging/structured/src/main/scala/tofu/logging/Logging.scala
```diff
@@ -65,13 +65,15 @@ trait LoggingBase[F[_]] {
 }
 
 /** Logging marked with Service type*/
-trait ServiceLogging[F[_], +Service] extends LoggingBase[F]
+trait ServiceLogging[F[_], Service] extends LoggingBase[F] {
+  final def to[Svc2]: ServiceLogging[F, Svc2] = this.asInstanceOf[ServiceLogging[F, Svc2]]
```
Z1kkurat: This will not change class of Logger that was created initially for `Service`. Is this a desired behaviour?

Issue #92, opened 2019-12-27
Implement whenOpt and unlessOpt


Odomontois:
Implement methods like `whenA` and `unlessA`, but returning `F[Option[A]]` in the `tofu.syntax.monadic`

PR #93, opened 2019-12-27
Solution to issue #92

FunFunFine:


higherKindCore/src/main/scala/tofu/syntax/monadic.scala
```diff
@@ -24,7 +24,12 @@ object monadic extends TupleSemigroupalSyntax with ApplicativeSyntax with MonadS
     def ap(fa: F[A])(implicit F: Apply[F]): F[B]  = F.ap(fab)(fa)
     def <*>(fa: F[A])(implicit F: Apply[F]): F[B] = F.ap(fab)(fa)
   }
-
+  impicit final class TofuApplicativeOps[F[_], A](private val fa: =>F[A]) extends AnyVal {
+    import scala.Option.{when, unless}
+    
+    def whenOpt(condition: Boolean)(implicit F: Applicative[F]): F[Option[()]]   = F.map(fa)(when(condition)(_))
```
Odomontois: First we'd like to completely avoid evaluation\applying effect of `fa` when condition is not met
Second, we'd like to have `F[Option[A]]` as a result

FunFunFine: Reworked this to be an extensions of condition so that fa is called by value.

PR #104, opened 2020-01-09
Deprecated limitedTraverse.

### LINKED ISSUE #103: Deprecate and/or remove tofu.concurrent.syntax.limitedTraverse
Cats-effect 2.0.0 already has this option in `Concurrent` along with corresponding syntax under `cats.effect.syntax.paralleln`: [link](https://github.com/typelevel/cats-effect/blob/dd8607baed11da140688d24e467ce76159517910/core/shared/src/main/scala/cats/effect/syntax/ParallelNSyntax.scala).  
There's no need in our implementation, we should probably deprecate and then remove (or just remove) it.

atapin:
Closes #103


Z1kkurat:
I think for now we can leave this as a deprecation warning and drop it in a couple of releases (since I know of users of this API, including myself)

Issue #109, opened 2020-01-13
Create summoners for specialized Readers


Odomontois:
Make it possible to summon specific contextual instances via 
```scala
HasContext[F, C]
HasLocal[F, C]    
HasContextRun[F, G, C]
```

DoSOfRedRiver:
I currently using a solution which works like:
```scala
  trait Id
  trait Templates

  type HasTemplates[F[_]] = HasContext[F, Templates]
  type HasId[F[_]] = HasContext[F, Id]

  def example[F[_]: HasTemplates: HasId: Monad] = {
    for {
      id <- hasContext[HasId].get
      templates <- hasContext[HasTemplates].get
    } yield ()
  }
```
Not sure if it is good enough, though. What do you think?

atapin:
@Odomontois should this issue be closed?

PR #110, opened 2020-01-13
Solution to #109

FunFunFine:
Made it possible to summon implicit instances from context


core/src/main/scala/tofu/package.scala
```diff
@@ -2,13 +2,25 @@ import cats.effect.Bracket
 import cats.{ApplicativeError, MonadError}
 
 package object tofu {
-  type HasContext[F[_], C]          = Context[F] { type Ctx    = C }
-  type HasLocal[F[_], C]            = Local[F] { type Ctx      = C }
+  type HasContext[F[_], C] = Context[F] { type Ctx = C }
+  object HasContext {
```
Odomontois: I fill those belong to Errors.scala

Z1kkurat: @odomontois Why do you think summoning instances belongs to Errors file?

Odomontois: package objects are nasty things in terms of refactoring, bytecode, binary compatibility
I'd like to have them as small as possible, so all the things, that are defineable in the package itself, should be defined in the package itself, I feel

FunFunFine:
finally moved summoners to Errors.scala as @Odomontois said

Z1kkurat:
@FunFunFine can you please format the files? Seems like our moder formatter doesn't quite like current formatting: https://travis-ci.com/TinkoffCreditSystems/tofu/jobs/282971940#L269

Z1kkurat:
LGTM. @Odomontois if you don't have any objections I will merge this

PR #115, opened 2020-01-15
fix ZLogs

road21:
Fix unusable trait `ZLogs`


zio/logging/src/main/scala/tofu/logging/package.scala
```diff
@@ -0,0 +1,30 @@
...
+
+    def withContext[R: Loggable]: ZLogs[R] = new ZLogs[R] {
+      def forService[Svc: ClassTag]: UIO[ZLogging[R]] =
+        UIO.effectTotal(new URIOZLoggingImpl(loggerForService[Svc]))
+      override def byName(name: String): UIO[ZLogging[R]] =
+        UIO.effectTotal(new URIOZLoggingImpl[R](LoggerFactory.getLogger(name)))
+    }
+  }
+
+  val ZLogs: ZLogsOps.type = ZLogsOps
```
Odomontois: That is unusual encoding, what does that fix?
Also why do we define `package object logging` in the module that doesn't own `package logging`
What if we decide to define `package object logging` in the logging core?
Would it be more wise define those in the `tofu.logging.zio` package?

road21: I just wanted to keep structure `tofu.logging.ZLogs`, but I totally agree that it will be much better to make separate `tofu.logging.zio`. Otherwise if we decide to define package object in core it will be unaccessible.
Fixed.

PR #122, opened 2020-01-22
`RestoreTo[F[_], G[_]]` extends `Lift[G[_], F[_]]`.

### LINKED ISSUE #117: Function allowing to get back into exceptional context in `RestoreTo`
Wouldn't it be useful to have a function allowing to get back from `G[_]` to `F[_]` in `RestoreTo`?

oskin1:
Close #117

### DIFF
--- a/optics/core/src/main/scala/tofu/optics/functions.scala
+++ b/optics/core/src/main/scala/tofu/optics/functions.scala
@@ -56,4 +56,6 @@ object functions {
   def some[A]: Subset[Option[A], A] = Subset[Option[A]](identity)(Some(_))
 
   def none[A]: Subset[Option[A], Unit] = Subset[Option[A]](_ => Some(()))(_ => None)
+
+  def extractAnyFromNothing: Extract[Nothing, Any] = (s: Nothing) => s
 }
--- a/optics/core/src/main/scala/tofu/optics/functions.scala
+++ b/optics/core/src/main/scala/tofu/optics/functions.scala
@@ -57,5 +57,5 @@ object functions {
 
   def none[A]: Subset[Option[A], Unit] = Subset[Option[A]](_ => Some(()))(_ => None)
 
-  def extractAnyFromNothing: Extract[Nothing, Any] = (s: Nothing) => s
+  def extractSubtype[A <: B, B]: Extract[A, B] = (s: B) => s
 }
--- a/zio/core/src/main/scala/tofu/zioInstances/ZioTofuInstance.scala
+++ b/zio/core/src/main/scala/tofu/zioInstances/ZioTofuInstance.scala
@@ -4,7 +4,7 @@ import cats.{Applicative, Functor, ~>}
 import cats.effect.{CancelToken, Fiber}
 import tofu.internal.CachedMatcher
 import tofu.lift.Unlift
-import tofu.optics.Contains
+import tofu.optics.{Contains, Extract}
 import tofu.syntax.functionK.funK
 import tofu.zioInstances.ZioTofuInstance.convertFiber
 import zio.clock.Clock
@@ -79,7 +79,7 @@ object ZioTofuInstance {
   }
 }
 
-class ZioTofuErrorsToInstance[R, E, E1](implicit lens: Contains[E1, E])
+class ZioTofuErrorsToInstance[R, E, E1](implicit lens: Extract[E1, E])
     extends ErrorsTo[ZIO[R, E, *], ZIO[R, E1, *], E] {
   final def handleWith[A](fa: ZIO[R, E, A])(f: E => ZIO[R, E1, A]): ZIO[R, E1, A] = fa.catchAll(f)
   final def restore[A](fa: ZIO[R, E, A]): ZIO[R, E1, Option[A]]                   = fa.option
--- a/zio/core/src/main/scala/tofu/zioInstances/ZioTofuInstance.scala
+++ b/zio/core/src/main/scala/tofu/zioInstances/ZioTofuInstance.scala
@@ -29,6 +29,7 @@ class ZioTofuInstance[R, E]
 
   final def restore[A](fa: ZIO[R, E, A]): ZIO[R, E, Option[A]] = fa.option
   final def raise[A](err: E): ZIO[R, E, A]                     = ZIO.fail(err)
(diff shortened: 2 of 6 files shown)


zio/core/src/main/scala/tofu/zioInstances/package.scala
```diff
@@ -17,7 +17,13 @@ class ZioInstances1 extends ZioInstances2 {
   private[this] val rioTofuInstanceAny: RioTofuInstance[Any] = new RioTofuInstance
   final def rioTofuInstance[R]: RioTofuInstance[R]           = rioTofuInstanceAny.asInstanceOf[RioTofuInstance[R]]
 
-  private[this] val zioErrorsToInstanceAny: ZioTofuErrorsToInstance[Any, Any, Nothing] = new ZioTofuErrorsToInstance
+  private[this] implicit val containsInstanceAny: Contains[Nothing, Any] =
+    new Contains[Nothing, Any] {
+      def set(s: Nothing, b: Any): Nothing = s
+      def extract(s: Nothing): Any = s.asInstanceOf[Any]
```
road21: asInstanceOf is redundant
Also, may be this instance should be somewhere in optics module?

oskin1: Maybe, but I haven't found more suitable place for such dummy implicit.

road21: I think such instance can be abstracted to smth like:
```scala
def subtype[A, B >: A]: Contains[A, B] = PContains[A, B](identity)((x, _) => x)
``` 
and moved to ```tofu.optics.functions```.
@Odomontois, what do you think?

Odomontois: I agree with moving, but I like the current non-allocating version more.

Odomontois: Also, I believe such instance would break Lens laws in general

zio/core/src/main/scala/tofu/zioInstances/ZioTofuInstance.scala
```diff
@@ -78,10 +79,12 @@ object ZioTofuInstance {
   }
 }
 
-class ZioTofuErrorsToInstance[R, E, E1] extends ErrorsTo[ZIO[R, E, *], ZIO[R, E1, *], E] {
+class ZioTofuErrorsToInstance[R, E, E1](implicit lens: Contains[E1, E])
```
Odomontois: It looks we need only `Extract` instance here

zio/core/src/main/scala/tofu/zioInstances/package.scala
```diff
@@ -17,7 +17,13 @@ class ZioInstances1 extends ZioInstances2 {
...
 
-  private[this] val zioErrorsToInstanceAny: ZioTofuErrorsToInstance[Any, Any, Nothing] = new ZioTofuErrorsToInstance
+  private[this] implicit val containsInstanceAny: Contains[Nothing, Any] =
+    new Contains[Nothing, Any] {
+      def set(s: Nothing, b: Any): Nothing = s
+      def extract(s: Nothing): Any = s
+    }
+
+  private[this] def zioErrorsToInstanceAny: ZioTofuErrorsToInstance[Any, Any, Nothing] = new ZioTofuErrorsToInstance[Any, Any, Nothing]
   final def zioTofuErrorsToInstance[R, E, E1]: ZioTofuErrorsToInstance[R, E, E1] =
```
Odomontois: This instance is not sound anymore.
You can not transform `ZIO[R, E1, A]` to `ZIO[R, E, A]` without any prior knowledge about how to map `E1` to `E`
I'm afraid trying to cast arbitrary `E1` to `Nothing` would bring `ClassCastException`

Odomontois: I suppose best we can do here is a change `E1` to `Nothing` explicitly for that instance.

zio/core/src/main/scala/tofu/zioInstances/package.scala
road21: Yep, sorry, my mistake.
Counterexample: for `Contains[Boolean, Boolean]` such instance, we have: ```extract(set(true, false)) = true```

zio/core/src/main/scala/tofu/zioInstances/package.scala
oskin1: @Odomontois, you mean `final def zioTofuErrorsToInstance[R, E]: ZioTofuErrorsToInstance[R, E, Nothing]`?

zio/core/src/main/scala/tofu/zioInstances/ZioTofuInstance.scala
```diff
@@ -78,10 +79,12 @@ object ZioTofuInstance {
   }
 }
 
-class ZioTofuErrorsToInstance[R, E, E1] extends ErrorsTo[ZIO[R, E, *], ZIO[R, E1, *], E] {
+class ZioTofuErrorsToInstance[R, E, E1](implicit lens: Extract[E1, E])
```
Odomontois: So now the only use of the `lens` parameter is to `Extract` something from `Nothing`
This is perfectly legit and could not touch performance (since mapError is never applied).
But it looks like we need at least one additional instance , e.g.
```scala
final def zioTofuExtractErrorsInstance[R, E, E1: * Extract E1]: ZioTofuErrorsToInstance[R, E, E1] = 
  new ZioTofuErrorsToInstance
```

To justify this encoding.

optics/core/src/main/scala/tofu/optics/functions.scala
```diff
@@ -56,4 +56,6 @@ object functions {
   def some[A]: Subset[Option[A], A] = Subset[Option[A]](identity)(Some(_))
 
   def none[A]: Subset[Option[A], Unit] = Subset[Option[A]](_ => Some(()))(_ => None)
+
+  def extractAnyFromNothing: Extract[Nothing, Any] = (s: Nothing) => s
```
Odomontois: Well, `Extract[A, B]` unlike `Lens[A, B]` is perfectly legal  for any `A <: B`, so it would be better to see more general case here.

PR #138, opened 2020-02-04
Add Context.const function that lifts value into a Context

pomadchin:
As an alternative to the [ApplicativeAsk.const](https://github.com/typelevel/cats-mtl/blob/9268fcb1469e53ba601bc765cf7d4f78e0b9e45a/core/src/main/scala/cats/mtl/ApplicativeAsk.scala#L36) from cats-mtl, it can be convenient to have a function that lifts values to the Context.

This idea and the code were brought by @optician


daron666:
👍

PR #152, opened 2020-02-12
Memo docs from community with love.

optician:


docs/memo.md
```diff
@@ -0,0 +1,189 @@
...
+## Dependency
+It is available in general package `"ru.tinkoff" %% "tofu" % Versions.tofu` and as standalone dependency `"ru.tinkoff" %% "tofu-memo" % Versions.tofu`
+
+## Functionality
+* A cache of a single value on access.
+* A cache of a values mapping on access.
+* TTL. Expired values are discarded on access.
+* Forced invalidation based on time.
+
+Missed:
```
Z1kkurat: I guess something like `Not implemented` would sound a bit better than `Missed`

docs/memo.md
```diff
@@ -0,0 +1,189 @@
...
+import cats.effect.{Clock, Sync}
+import cats.syntax.applicative._
+import cats.syntax.flatMap._
+import cats.syntax.functor._
+import monix.eval.Task
+import monix.execution.Scheduler.Implicits.global
+import scala.concurrent.duration.FiniteDuration
+import java.util.concurrent.TimeUnit
+
+def effect[F[_] : Sync]: F[Int] = Sync[F].delay(println("called")) >> 335.pure[F]
```
Z1kkurat: Is `>> 335.pure[F]` intentional? Since you have `cats.syntax.functor._` import, you could write that like `Sync[F].delay(println("called")).as(335)`

docs/memo.md
```diff
@@ -0,0 +1,189 @@
...
+
+/*
+called
+Number 335
+Number 335
+called
+Number 335
+ */
+```
+How it works. Cached value remembers time of read access during update (T_update). On any read access tofu selects a time to compare with T_update (update value when `T_selected > T_update`). Selection is based on a formula `max(T_now - TTL, T_CacheControl)`. If you set `T_CacheControl = T_now` then always T_CacheControl is more than T_update. After update of a value, T_update equals T_CacheControl.
```
Z1kkurat: Would be nice to emphasize the name of the block to make it more distinguishable visually from other text, something like header would do.
Also, not sure about `tofu selects` part. Library itself doesn't `select` anything, it's an implementation that makes any work.

docs/memo.md
optician: Sorry but dont see a point.

optician: I dont like `not implemented` at all. It is like a promiss. Changed to neutral.

docs/memo.md
optician: Changed.

docs/memo.md
optician: This text is a part of the sample. Why it have be separated?

> Library itself doesn't select anything
It's about selection of a computational branch.

PR #164, opened 2020-02-21
Add logging documentation

### LINKED ISSUE #108: Add docs for logging module
I’m quite annoyed by constant complaining in Scala Group about tofu documentation. Let’s change that, starting from logging.

Z1kkurat:
This closes #108


docs/logging.md
```diff
@@ -0,0 +1,160 @@
...
+  def doLogging: F[Unit] =
+    for {
+        _ <- info"This is a syntax demo: $user"
+    } yield () 
+}
+
+object MyService {
+  // note, that creating service may require Sync instance, but class itself does not!
+  def apply[F[_]: Sync]: F[MyService[F]] = {
+    val logs = Logs.sync[F, F]
```
Odomontois: I would not recommend creating `Logs` per service.
It more like `Logs` per application, so Service initialization would require `Logs[F, F]`

docs/logging.md
```diff
@@ -0,0 +1,160 @@
+---
+id: logging
+title: Logging
+---
+
+### Logging
+
+Almost every application nowadays does logging, whether it's logging of business events, history of some state of 
+application or even debug record sequences. We log to files, distributed systems, standard outputs of OSes, and we 
+got used to it so much it seems really simply to us. Many existing libraries provide many ways of logging from
```
happy-bracket: really simple*

docs/logging.md
```diff
@@ -0,0 +1,160 @@
...
+### Logging
+
+Almost every application nowadays does logging, whether it's logging of business events, history of some state of 
+application or even debug record sequences. We log to files, distributed systems, standard outputs of OSes, and we 
+got used to it so much it seems really simply to us. Many existing libraries provide many ways of logging from 
+your application's code, either automatic or manual.
+
+### What's wrong?
+
+First, in fact, logging is a side effect. Whether we log anything to a standard output, or a file - it's an interaction
```
happy-bracket: Unclear as to whether "in fact" relates to the subsequent statement "logging is a side effect" or it is stylistically incorrect use of transitional phrases.
Assuming the former, I would recommend recomposing this sentence as "First, logging is, in fact, a side effect"

docs/logging.md
```diff
@@ -0,0 +1,160 @@
...
+application or even debug record sequences. We log to files, distributed systems, standard outputs of OSes, and we 
+got used to it so much it seems really simply to us. Many existing libraries provide many ways of logging from 
+your application's code, either automatic or manual.
+
+### What's wrong?
+
+First, in fact, logging is a side effect. Whether we log anything to a standard output, or a file - it's an interaction 
+with the outside world ('outside' relative to our program). Thus, logging as a process can be represented as an effect and operations - as an algebra.  
+Second, logging itself seems easy, but stuff around it - not so much. MDC, contexts, structured logging - all these 
+things are very helpful on a day-to-day basis, helping us to read and write our logs, making them clean, comprehensible
```
happy-bracket: "helpful... helping"
Slightly disturbing repetition, maybe replace with a synonim?
From the top of my head: "helpful" -> "handy", or "helping" -> "assisting", "aiding"

docs/logging.md
Z1kkurat: Thanks, I made an example a bit clearer with notes

docs/logging.md
```diff
@@ -0,0 +1,167 @@
...
+Second, logging itself seems easy, but stuff around it - not so much. MDC, contexts, structured logging - all these 
+things are very helpful on a day-to-day basis, allowing us to read and write our logs, making them clean, comprehensible 
+and useful for many purposes. But there are few simple ways to provide all this functionality without bleeding and 
+screaming from seeing over-engineered solutions.
+
+### The solution: Logging
+
+Tofu offers an easy and understandable abstraction over logging (it's even named like this!) but keeping in mind that 
+logging is an effect that can be composed.  
+It looks very similar to interfaces we all are used to (from SLF4J and other logging libraries and facades), introducing some new concepts:
```
happy-bracket: The conclusion of the paragraph is almost identical, the only difference is clarifying what actually is the new feature. Maybe collapse it with this sentence? That would additionally provide the reader with an information of what new concepts to look for, compared to "old" interfaces

docs/logging.md
```diff
@@ -0,0 +1,167 @@
...
+
+// for simplification, you can use whatever F you like
+type F[A] = cats.effect.IO[A]
+
+class MyService
+
+// defines simple logging behaviour without any notion of contexts
+val syncLogs = Logs.sync[F, F]
+
+// desribing creation of logging instances and logging itself. When being run, this will log a message
```
happy-bracket: desribing -> describing

docs/logging.md
```diff
@@ -0,0 +1,167 @@
...
+effect type and a `Loggable` for your context. 
+That's natural - describe how we can obtain a context from a computation `F` and a way to represent that context in log, 
+and we will ensure it will be logged anytime you use `Logging` operations. 
+You can log that context to string as well as use it as a part of structured logging. 
+
+#### What LoggedValue and Loggable is
+`LoggedValue` is a representation of a value that can be written to a log in a string form. 
+It's used internally, and you do not need to use it directly most of the time. Instead, what you want is a `Loggable` 
+typeclass, that describes how a value of some type `A` can be logged, either as a string representation in log message 
+and as a component of structured logging. Given an instance of `Loggable` for a type, a value of the type can be converted
```
happy-bracket: "and" is prefaced with "either", however, "either" should be followed by "or", since it's a choice construct

logging/structured/src/main/scala/tofu/logging/Logs.scala
```diff
@@ -12,8 +12,34 @@ import tofu.syntax.monadic._
...
+  * }}}
+  */
 trait Logs[+I[_], F[_]] extends LogsVOps[I, F] {
+
+  /** Creates an instance of [[tofu.logging.Logging]] with a given arbitrary type tag.
+    * In a simple way that means creating a Logger for a given class. */
   def forService[Svc: ClassTag]: I[Logging[F]]
+
+  /** Creates an instance of [[tofu.logging.Logging]] for a given artitrary string name.
+    *  In a simple way that means creating a Logger with a given name.
```
optician: The difference between the two sentences is too small.

logging/structured/src/main/scala/tofu/logging/Logs.scala
```diff
@@ -48,13 +78,24 @@ object Logs {
...
     def byName(name: String): I[Logging[F]]      = logging.pure[I]
   }
 
+  /**
+    * Retuns an instance of [[tofu.logging.Logs]] that will produce a no-op [[tofu.logging.Logging]] instances.
+    */
   def empty[I[_]: Applicative, F[_]: Applicative]: Logs[I, F] = const[I, F](Logging.empty[F])
 
+  /**
+    * Combines two instances of [[tofu.logging.Logs]], resulting in a single one that will produce
```
optician: > Combines to produce instances by combining. 

It looks useless to me. Imho it's better to tell about the behavior of the result.

docs/logging.md
```diff
@@ -0,0 +1,165 @@
...
+```scala mdoc
+import tofu.logging._
+
+// for simplification, you can use whatever F you like
+type F[A] = cats.effect.IO[A]
+
+class MyService
+
+// defines simple logging behaviour without any notion of contexts
+val syncLogs = Logs.sync[F, F]
```
optician: Is it a 'heavy' operation?

docs/logging.md
```diff
@@ -0,0 +1,165 @@
...
+
+There are multiple ways to create `Logs` instance, among them:
+* `Logs.sync`, requiring a `cats.effect.Sync` instance for delaying side-effects
+* `Logs.withContext`, requiring a `cats.effect.Sync` instance for delaying side-effects and a proof that `F` has a context to log
+* `Logs.const` that wraps single `Logging` thus always returning the same instance
+* `Logs.empty` that will produce a no-op `Logging` instances
+* `Logs.combine` that can combine two instances of `Logs` thus combining produces instances
+
+
+#### Context logging
```
optician: Imho it would be better to swap "Context Logging" and "What LoggedValue and Loggable is".

docs/logging.md
```diff
@@ -0,0 +1,165 @@
...
+It is possible to log the context of your computations (MDC, remember?) if you have an instance of `tofu.Context` for your
+effect type and a `Loggable` for your context. 
+That's natural - describe how we can obtain a context from a computation `F` and a way to represent that context in log, 
+and we will ensure it will be logged anytime you use `Logging` operations. 
+You can log that context to string as well as use it as a part of structured logging. 
+
+#### What LoggedValue and Loggable is
```
optician: Maybe it would be more descriptive to call the chapter something like "Logs Representation" or "Logs Format".

docs/logging.md
```diff
@@ -0,0 +1,165 @@
...
+`LoggedValue` is a representation of a value that can be written to a log in a string form. 
+It's used internally, and you do not need to use it directly most of the time. Instead, what you want is a `Loggable`
```
optician: > It's used internally, and you do not need to use it directly most of the time.

So why do we start talking about `LoggedValue`?

docs/logging.md
```diff
@@ -0,0 +1,165 @@
...
+  }
+  
+  override def logShow(a: User): String = s"name = ${a.name}, surname = ${a.surname}"
+}
+```
+
+Here, we defined our `Loggable` as a `DictLoggable` that has a notion of `fields`. 
+Fields are a structure of your `Loggable` that is used for structured logging by your `LogRendered`.
+Tofu comes with a `tofu.logging.ElkLayout` (present in `tofu-logging-layout` library), that will log the structure (fields) of your `Loggable`
+as a JSON fields.
```
optician: **a** JSON **fields**

docs/logging.md
Z1kkurat: Because user API showcases functions like `def trace(message: String, values: LoggedValue*)` and it may be unclear why do we define `Loggable` instances when API requires `LoggedValue`s

docs/logging.md
Z1kkurat: No it's not. But we advice users to share an instance of `Logs`, maybe keep a single one for a whole application.

docs/logging.md
```diff
@@ -0,0 +1,165 @@
+---
+id: logging
+title: Logging
+---
+
+### Logging
+
+Almost every application nowadays does logging, whether it's logging of business events, history of some state of
```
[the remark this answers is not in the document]
Z1kkurat: @Odomontois WDYT? Is is appropriate here or should I remove it?

docs/logging.md
optician: As for me it needs a clarification.

docs/logging.md
optician: > Because user API showcases functions 

API is separated from the sentence by subtitle. Context is lost.

I suggest teaching a user how to
* log objects
* format logs
and not what some classes/traits can do. Think about it like inversion of accents.

docs/logging.md
```diff
@@ -0,0 +1,168 @@
...
+
+#### Loggable example
+```scala mdoc
+import tofu.logging._
+
+case class User(name: String, surname: String)
+
+implicit val userLoggable = new DictLoggable[User] {
+  override def fields[I, V, R, S](a: User, i: I)(implicit r: LogRenderer[I, V, R, S]): R = {
+    r.addString("name", a.name, i)
```
Odomontois: This is an incorrect implementation. 
It would work for mutable renderers only.
Any warning generator should warn you about illegal `R` dropping.
You mus use `|+|` combinator to compose multiple fields addition

PR #165, opened 2020-02-22
Add MakeRef documentation

valery-shinkevich:


concurrent/src/main/scala/tofu/concurrent/MakeRef.scala
```diff
@@ -3,21 +3,61 @@ package tofu.concurrent
...
   def apply[F[_]](implicit agents: Refs[F]): MakeRef.Applier[F, F] = new MakeRef.Applier[F, F](agents)
 }
 
 object MakeRef {
+
+  /**
+    * Makes a `Ref` instance
+    *
+    * Uses the
+    * [[https://tpolecat.github.io/2015/07/30/infer.html Kinda-Curried Type Parameters]]
```
Z1kkurat: This blogpost is a bit outdated. There is a description of this technique in cats docs (which mentions Rob's post in turn): https://typelevel.org/cats/guidelines.html#partially-applied-type-params

valery-shinkevich: Yes, but, IMHO, this post is more complicated and uses the Applicative [F]. Ok, I will change link.

PR #169, opened 2020-02-25
Add convenient syntax for working with F[Either[E, A]]

### LINKED ISSUE #149: Add convenient syntax for working with F[Either[E, A]]
We interested in few handy methods, that would allow to do basic things with `F[Either[E, A]]` with additional context `Monad[F]` without wrapping it in `EitherT`
possible list is
```scala
def orElse(f: => F[Either[E, A]]): F[Either[E, A]]
def catchAll(e: E => F[A]): F[A]
def absolve(implicit R: Raise[F, E]): F[A]
def flatMapR[B](f: A => F[B]): F[Either[E, B]]
def flatMapL[E](f: E => F[E1]): F[Either[E1, A]]
def doubleFlatMap[B](f: A => F[Either[E, B]]): F[Either[E, B]]
```

Ssstlis:
This PR resolve #149
I had some questions about documentation in source file and do i need to use helpers in tests or not?


core/src/main/scala/tofu/syntax/feither.scala
```diff
@@ -0,0 +1,210 @@
...
+
+    def mapK[G[_]](implicit F: ~>[F, G]): G[Either[L, R]] = {
+      F(e)
+    }
+
+    def liftTo[G[_]](implicit F: Lift[F, G]): G[Either[L, R]] = {
+      F.lift(e)
+    }
+
+    def mergeF(implicit ev: L =:= R, F: Functor[F]): F[R] = {
```
road21: ```def merge[A >: R](implicit ev: L <:< A, F: Functor[F]): F[A]```?

core/src/main/scala/tofu/syntax/feither.scala
```diff
@@ -0,0 +1,210 @@
...
+      F.map(productF(eb))(_.map { case (r, r1) => f(r, r1) })
+    }
+
+    def flatMap2F[L1 >: L, R1, Z](
+        eb: => F[Either[L1, R1]]
+    )(f: (R, R1) => F[Z])(implicit F: Monad[F]): F[Either[L1, Z]] = {
+      F.flatMap(productF(eb))(_.traverse { case (r, r1) => f(r, r1) })
+    }
+
+    def mapK[G[_]](implicit F: ~>[F, G]): G[Either[L, R]] = {
```
road21: I don't think that's a good idea to make natural transformation implicit.
Do we really need mapK and liftTo methods here?

core/src/main/scala/tofu/syntax/feither.scala
```diff
@@ -0,0 +1,210 @@
...
+        case Left(left)       => f(left)
+        case right @ Right(_) => F.pure(right.leftCast)
+      }
+    }
+
+    def doubleFlatMap[L1 >: L, R1 >: R](f: R => F[Either[L1, R1]])(implicit F: Monad[F]): F[Either[L1, R1]] = {
+      F.flatMap(e) {
+        case Right(right)   => f(right)
+        case left @ Left(_) => F.pure(left.rightCast)
+      }
```
road21: I think we can keep just one method from doubleFlatMap, flatMapF

core/src/main/scala/tofu/syntax/feither.scala
Ssstlis: At the last minute i became doubted about it but decided to left it. So, if it doesn't looks good enough i can cut it.

core/src/main/scala/tofu/syntax/feither.scala
Ssstlis: Ok, seems good

PR #186, opened 2020-03-05
Promotion of lenses of marked inner fields on classy lenses generation

Odomontois:


optics/macro/src/test/scala-2.13/tofu/optics/macros/TestClassyContains.scala
```diff
@@ -4,11 +4,25 @@ import tofu.optics.{Contains, Label, PContains, Update}
 object TestClassyContains {
   // compile test for searching classy optics
   implicitly[Contains[FooBar2, Int] with Label["i"]]
+  implicitly[Contains[FooBar2, Int] with Label["i"]]
```
danslapman: Seems to be an unnecessary duplication

PR #188, opened 2020-03-05
Provide and HasProvide

road21:


core/src/main/scala/tofu/Context.scala
```diff
@@ -44,10 +44,24 @@ object Local extends LocalInstances[λ[(f[_], g[_], r) => HasLocal[f, r]]] {
 
 trait LocalInstances[TCA[f[_], g[_], r] >: HasLocal[f, r]] extends RunContextInstances[TCA]
 
-trait RunContext[F[_]] extends Local[F] {
```
Odomontois: I suppose we need to keep at least deprecated alias for this

road21: I haven't drop it, just added one more parent

PR #191, opened 2020-03-06
Add higher kinded typeclass instances for Logs

catostrophe:
1. Removed the `Embed[Logging]` instance overlapped by the `RepresentableK[Logging]` instance and thus unsummonable.
2. Added `MonoidalK` instances for `Logs` for the second type param.

### DIFF
--- a/logging/structured/src/main/scala/tofu/logging/Logs.scala
+++ b/logging/structured/src/main/scala/tofu/logging/Logs.scala
@@ -96,7 +96,7 @@ trait Logs2FunctorK[Y[_]] extends FunctorK[Logs[Y, *[_]]] {
 
   def mapK[F[_], G[_]](af: Logs[Y, F])(fk: F ~> G): Logs[Y, G] = new Logs[Y, G] {
     def forService[Svc: ClassTag]: Y[Logging[G]] = af.forService[Svc].map(_.mapK(fk))
-    def byName(name: String): Y[Logging[G]] = af.byName(name).map(_.mapK(fk))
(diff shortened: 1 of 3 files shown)


logging/structured/src/main/scala/tofu/logging/Logs.scala
```diff
@@ -73,3 +80,44 @@ object Logs {
...
+trait Logs2FunctorK[Y[_]] extends FunctorK[Logs[Y, *[_]]] {
+  implicit def I: Functor[Y]
+
+  def mapK[F[_], G[_]](af: Logs[Y, F])(fk: F ~> G): Logs[Y, G] = new Logs[Y, G] {
+    def forService[Svc: ClassTag]: Y[Logging[G]] = af.forService[Svc].map(_.mapK(fk))
+    def byName(name: String): Y[Logging[G]]      = af.byName(name).map(_.mapK(fk))
+  }
+}
+
+trait Logs2ApplyK[Y[_]] extends Logs2FunctorK[Y] with ApplyK[Logs[Y, *[_]]] {
```
Odomontois: Is this necessary to make them by hand?
tofu-logging depends on cats-tagless-macros, so one can `Derive` them

catostrophe: Afaik, cats.tagless.Derive can't derive such instances. I tried but didn't manage to. Will double-check

Odomontois: Sorry. My Bad.

PR #195, opened 2020-03-11
Fix zio interop instances and implicits

catostrophe:
1. Small fixes
2. Added implicits for `Unlift`


optics/core/src/main/scala/tofu/optics/functions.scala
```diff
@@ -58,4 +58,9 @@ object functions {
   def none[A]: Subset[Option[A], Unit] = Subset[Option[A]](_ => Some(()))(_ => None)
 
   def extractSubtype[A <: B, B]: Extract[A, B] = (s: B) => s
+
+  def containsUnit[A, B >: Unit]: Contains[A, B] = new Contains[A, B] {
```
Odomontois: that's smart

PR #204, opened 2020-03-14
More fs2 instances for Context hierarchy

catostrophe:


fs2/src/main/scala/tofu/fs2Instances/package.scala
```diff
@@ -1,11 +1,3 @@
...
-import fs2._
-
-package object fs2Instances {
-  private[this] val fs2InstanceAny = new FS2StreamInstance[Any]
-
-  final implicit def fs2StreamInstance[A]: FS2StreamInstance[A] = fs2InstanceAny.asInstanceOf[FS2StreamInstance[A]]
-  final implicit def fs2StreamContext[F[_], R](implicit fctx: F HasContext R): HasContext[Stream[F, *], R] =
-    new FS2Context[F, R]
-}
+package object fs2Instances extends Fs2Instances1
```
gurinderu: Maybe it is better to rename this object to `instances` and place them under the fs2 package?

catostrophe: I didn't change the existing naming, but I'm for some changes here. I agree with @gurinderu that `tofu.fs2.instances._` and `tofu.fs2.implicits._` would look better than the current naming. I'd also like to refactor `zioInstances` into 2 objects correspondingly.

PR #215, opened 2020-04-03
Fix derivation of UnliftIO[Env[E, *]]

catostrophe:
This is to make UnliftIO instance more derivable


catostrophe:
Pls review @Odomontois @Z1kkurat

Issue #94, opened 2019-12-27
Fix TofuBracketMVarOps.bracketUpdate


Odomontois:
Current method is nonsense, we should transform it to a method, accepting `A => F[A]` and putting result on success,

Behavior should be related to datatype described in #95

PR #241, opened 2020-04-23
Return from bracketUpdate only on successful update

mehakun:
Fixes issue #94 
Maybe old behaviour should be preserved and new method should be created with name like `bracketUpdateStrong`

### DIFF
--- a/core/src/main/scala/tofu/syntax/bracket.scala
+++ b/core/src/main/scala/tofu/syntax/bracket.scala
@@ -36,8 +36,15 @@ object bracket {
   }
 
   implicit final class TofuBracketMVarOps[F[_], A](private val mvar: MVar[F, A]) extends AnyVal {
+
+    /**
+      * Update value with effectful transformation, wait for the new result.
+      * Works similarly to [[tofu.concurrent.Agent.updateM]]
+      * @param use function to atomically modify value contained in `Bracket`
+      * @return `F[A]` modified value contained in `Bracket`
+      */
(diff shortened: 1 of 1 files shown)


core/src/main/scala/tofu/syntax/bracket.scala
```diff
@@ -36,8 +36,15 @@ object bracket {
...
+
+    /**
+      * Update value with effectful transformation, wait for the new result.
+      * Works similarly to [[tofu.concurrent.Agent.updateM]]
+      * @param use function to atomically modify value contained in `Bracket`
+      * @return `F[A]` modified value contained in `Bracket`
+      */
     def bracketUpdate[B, E](use: A => F[B])(implicit F: Applicative[F], FG: Guarantee[F]): F[B] =
-      mvar.take.bracketIncomplete(use)(mvar.put)
+      mvar.take.bracketOpt(use)((b, isUpdated) => mvar.put(b).whenA(isUpdated))
```
Odomontois, line 47: So what if `isUpdated == false` bracket will remain empty?

mehakun: Yes it will remain empty. Is expected semantic to always put value into bracket?

mehakun: I guess always putting value into mvar is the most flexible way for users. That way they can choose between retrying operation or performing it once

Odomontois:
Closing for now

PR #253, opened 2020-05-12
Make Timeout instance for Concurrent types implicit

FunFunFine:
This method is not referenced anywhere and it seems like `implicit` keyword is missing.

Current workaround is to write `implicit val timeoutInstance: Timeout[F] = Timeout.concurrent` and this doesn't look right

### DIFF
--- a/core/src/main/scala/tofu/Timeout.scala
+++ b/core/src/main/scala/tofu/Timeout.scala
@@ -4,6 +4,7 @@ import cats.effect.{Concurrent, ContextShift, IO, Timer}
 import simulacrum.typeclass
 import cats.syntax.apply._
 import cats.syntax.functor._
+import tofu.internal.NonTofu
 
 import scala.concurrent.duration.FiniteDuration
 
@@ -13,7 +14,7 @@ trait Timeout[F[_]] {
 }
 
 object Timeout {
-  implicit def concurrent[F[_]](implicit F: Concurrent[F], timer: Timer[F]): Timeout[F] =
+  implicit def concurrent[F[_]](implicit F: Concurrent[F], timer: Timer[F], nonTofu: NonTofu[F]): Timeout[F] =
     new Timeout[F] {
       override def timeoutTo[A](fa: F[A], after: FiniteDuration, fallback: F[A]): F[A] =
         F.race(fa, timer.sleep(after) *> fallback).map(_.merge)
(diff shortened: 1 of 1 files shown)


core/src/main/scala/tofu/Timeout.scala
```diff
@@ -13,7 +13,7 @@ trait Timeout[F[_]] {
 }
 
 object Timeout {
-  def concurrent[F[_]](implicit F: Concurrent[F], timer: Timer[F]): Timeout[F] =
+  implicit def concurrent[F[_]](implicit F: Concurrent[F], timer: Timer[F]): Timeout[F] =
```
Odomontois, line 16: Can you please add `NonTofu` requirement as it is here https://github.com/TinkoffCreditSystems/tofu/blob/8d9b7b846d7e5109b5055e64b439ad4b6ac52d90/core/src/main/scala/tofu/Fire.scala#L33

Odomontois:
This method was intended to use explicitly to define your own instance.
But I don't see big harm to make it implicit with proper safeguard

core/src/main/scala/tofu/Timeout.scala
FunFunFine: Done

core/src/main/scala/tofu/Timeout.scala
```diff
@@ -13,7 +14,7 @@ trait Timeout[F[_]] {
 }
 
 object Timeout {
-  def concurrent[F[_]](implicit F: Concurrent[F], timer: Timer[F]): Timeout[F] =
+  implicit def concurrent[F[_]](implicit F: Concurrent[F], timer: Timer[F], nonTofu: NonTofu[F]): Timeout[F] =
     new Timeout[F] {
       override def timeoutTo[A](fa: F[A], after: FiniteDuration, fallback: F[A]): F[A] =
         F.race(fa, timer.sleep(after) *> fallback).map(_.merge)
```
catostrophe, line 20: Isn't related to this change, but shouldn't it be a race between `fa` and `sleep`?
Something like: `F.race(timer.sleep(after), fa).getOrElseF(fallback)`?

Odomontois: Yeah definitely. @FunFunFine can you include that fix as well?

Odomontois: Will lock it with "request changes for now"

PR #259, opened 2020-05-13
Use ContravariantRaise instead of Raise in syntax.foption and syntax.feither

FunFunFine:
This allows one to use changed methods with ContravariantRaise in scope only.

### DIFF
--- a/core/src/main/scala/tofu/control/ContravariantFilter.scala
+++ b/core/src/main/scala/tofu/control/ContravariantFilter.scala
@@ -12,5 +12,5 @@ trait ContravariantFilter[F[_]] extends Contravariant[F] with Optional[F] {
   override def optional[A](fa: F[A]): F[Option[A]] = contramapFilter(fa)(identity)
 
   def contraFilter[A](fa: F[A])(f: A => Boolean): F[A] =
-    contramapFilter(fa)(a => if (f(a)) Some(a) else None)
+    contramapFilter(fa)(a => Some(a).filter(f))
 }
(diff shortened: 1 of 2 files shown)


core/src/main/scala/tofu/syntax/foption.scala
```diff
@@ -25,7 +25,7 @@ object foption {
         case x    => x.pure[F]
       }
 
-    def orThrow[E](err: => E)(implicit F: Monad[F], FE: Raise[F, E]): F[A] =
+    def orThrow[E](err: => E)(implicit F: Monad[F], FE: ContravariantRaise[F, E]): F[A] =
```
catostrophe, line 28: This change makes this ops method **less** universal than it is.

With the current implementation you still can write this:
```scala
  import cats.syntax.all._
  import tofu.syntax.foption._
  def test[F[_] : Monad : Raise[*[_], CommonError]]: F[Unit] = {
    ().some.pure[F].orThrow(ConcreteError())
  }
```
having `Upcast[CommonError, ConcreteError]` in scope. This works due to `raiseUpcast` implicit def. And `ConcreteError` **doesn't need** to be a subclass of `CommonError`. 

If you change the constraint to `ContravariantRaise`, you'll get this code working with direct subclasses via contravariance on E, but lose implicit derivation chain via `raiseUpcast` and the code above will break if `ConcreteError` isn't a subclass of `CommonError` (which is a rare case though).

If you need this change, I would insist on adding extra implicit derivations to make the code above work.

core/src/main/scala/tofu/control/ContravariantFilter.scala
```diff
@@ -12,5 +12,5 @@ trait ContravariantFilter[F[_]] extends Contravariant[F] with Optional[F] {
   override def optional[A](fa: F[A]): F[Option[A]] = contramapFilter(fa)(identity)
 
   def contraFilter[A](fa: F[A])(f: A => Boolean): F[A] =
-    contramapFilter(fa)(a => if (f(a)) Some(a) else None)
+    contramapFilter(fa)(a => Some(a).filter(f))
```
Ssstlis, line 15: I don't think this change simplify code, but also lose performance

Odomontois:
Since we agreed on another solution. This goes off.

PR #272, opened 2020-05-22
Add doobie module

catostrophe:
Added some useful stuff to make it easier to work with `doobie` when the main business-logic effect F is contextual and the context is used in the database layer (e.g. for logging and tracing).

Added an example.


gurinderu:
Amazing. Works perfectly in my per project

oskin1:
@catostrophe, let's add syntax for `LiftConnectionIO` in order to be able to lift ConnectionIO value like `(q: ConnectionIO[Int]).lift[DB]`?

catostrophe:
> @catostrophe, let's add syntax for `LiftConnectionIO` in order to be able to lift ConnectionIO value like `(q: ConnectionIO[Int]).lift[DB]`?

This should be added to tofu.syntax.lift._

catostrophe:
> @catostrophe, let's add syntax for `LiftConnectionIO` in order to be able to lift ConnectionIO value like `(q: ConnectionIO[Int]).lift[DB]`?

Could you make a PR to enhance lift syntax?

catostrophe:
> > @catostrophe, let's add syntax for `LiftConnectionIO` in order to be able to lift ConnectionIO value like `(q: ConnectionIO[Int]).lift[DB]`?
> 
> Could you make a PR to enhance lift syntax?

#283

doobie/src/main/scala/tofu/doobie/transactor/Txr.scala
```diff
@@ -0,0 +1,118 @@
...
+
+  /** Translates the stream, applying the transactional strategy. */
+  def transP: Stream[DB, *] ~> Stream[F, *]
+
+  /** Translates the stream, no strategy applied. */
+  def rawTransP: Stream[DB, *] ~> Stream[F, *]
+}
+
+object Txr {
+  type Aux[F[_], DB0[_]]     = Txr[F] { type DB[x] = DB0[x] }
```
oskin1, line 46: @catostrophe , I'd propose to swap F[_] and DB0[_] type parameters here in order to give Aux[F, DB] more intuitive look as transactor essentially provides a set of transformations from DB[_] to F[_] (Aux[DB, F]).

catostrophe: `Aux` pattern is widely used and the common way of defining an `Aux` alias is the following: type params go first, type members go last.
I don't think it's correct to swap the order just because there's `DB ~> F` transformation. What if I added the opposite one?

oskin1: @catostrophe, you mean as an alias like `type Trans[DB[_], F[_]] = Txr.Aux[F, DB]`?
Yes, I think this would be fine.

catostrophe: No, I meant why to choose the order of params in `Aux` based on the direction of arrows? There might be two opposite arrows. And what then? See `cats.Parallel.Aux`.
I'm just saying that `Aux[F, DB0]` is the common way of defining an aux.

catostrophe: What do you think of renaming `Txr` to `Trans` btw?

oskin1: @catostrophe, good idea, Trans is more intuitive name

catostrophe: @oskin1 Hmm... I remembered why I chose a different name.

```scala
val trans  = Trans.contextual[F](transactor)
...
val res = dbProgram ||> trans.trans
```
I thought `trans.trans` might be confusing.

Compare with:
```scala
val txr  = Txr.contextual[F](transactor)
...
val res = dbProgram ||> txr.trans
```
TXN is a commonly used acronym for transaction. So I chose TXR for transactor.

oskin1: Ok, makes sense

Odomontois:
I'm merging this blindly having good faith in your intentions.

PR #276, opened 2020-05-24
Auto-summonable optics instances for `Errors`

oskin1:


oskin1:
> Could you also make Subset[E, E1] auto-summonable?

@catostrophe , np, BTW what are the use cases of `Subset` in comparison with `Upcast`?

catostrophe:
> > Could you also make Subset[E, E1] auto-summonable?
> 
> @catostrophe , what are the use cases where `Upcast[E, E1]` instance is not enough and `Subset[E, E1]` is required?
> 
> It seems that whenever we can auto-derive `Sebset[E, E1]` for `E1 <: E` we can also derive `Upcast[E, E1]`, so why do we need `Sebset` auto-summoning?

`Subset` is just a prism. It is used in error handling. Why don't we have an autosummonable Subset instance for ADTs yet?

catostrophe:
> `Subset` is just a prism. It is used in error handling. Why don't we have an autosummonable Subset instance for ADTs yet?

And we cannot share `Subset` and `Upcast` instances because the former requires `ClassTag` for patmatching on error subtype.

core/src/test/scala/tofu/examples/TofuRaiseExamples.scala
```diff
@@ -0,0 +1,84 @@
+package tofu.examples
```
catostrophe, line 1: I suggest narrowing package namespace. Otherwise classes from different examples may clash.

oskin1: @catostrophe, can you be more specific?

catostrophe: @oskin1 Sorry. This doesn't seem to be an issue. I meant classes in the `tofu.examples` package in different modules might clash.

PR #279, opened 2020-05-25
Bracket mvar opts agent behaviour

mehakun:
Replace almighty and error-prone `bracketUpdate(A => F[B]):F[B]` with `bracketUpdate(A => F[A]):F[A]` and `bracketModify(A => F[(A, B)]):F[B]`.  
Also fixed bug in Applicative syntax in `unless_` method. It's implementation was same as `when_`


core/src/main/scala/tofu/syntax/bracket.scala
```diff
@@ -36,8 +36,31 @@ object bracket {
...
+      M.flatMap(mvar.take)(oldA =>
+        M.pure(oldA).bracketOpt(use) { (newA, success) => mvar.put(if (success) newA else oldA) }
+      )
+
+    /**
+      * Modify value with effectful transformation, calculating result. In case of error value remains unchanged
+      * @param use function to atomically modify value contained in `MVar` and produce result
+      * @return `F[B]`
+      */
+    // NOTE: maybe return F[(A, B)] to inspect updated mvar?
```
mehakun, line 55: notice the note :)

core/src/main/scala/tofu/syntax/bracket.scala
```diff
@@ -36,8 +36,31 @@ object bracket {
...
-      mvar.take.bracketIncomplete(use)(mvar.put)
+
+    /**
+      * Update value with effectful transformation. In case of error value remains unchanged
+      * @param use function to atomically modify value contained in `MVar`
+      * @return `F[A]` modified value contained in `MVar`
+      */
+    def bracketUpdate(use: A => F[A])(implicit FG: Guarantee[F], M: Monad[F]): F[A] =
+      M.flatMap(mvar.take)(oldA =>
+        M.pure(oldA).bracketOpt(use) { (newA, success) => mvar.put(if (success) newA else oldA) }
```
Odomontois, line 47: Looks like we need `guaranteeOpt` syntax in the `bracket` package

core/src/main/scala/tofu/syntax/bracket.scala
Odomontois: If user wants some information from `A` they obviously may use appropriate `B`.

Odomontois:
Ok will merge after `NOTE:` deletion

core/src/main/scala/tofu/syntax/bracket.scala
```diff
@@ -36,8 +36,31 @@ object bracket {
...
-    def bracketUpdate[B, E](use: A => F[B])(implicit F: Applicative[F], FG: Guarantee[F]): F[B] =
-      mvar.take.bracketIncomplete(use)(mvar.put)
+
+    /**
+      * Update value with effectful transformation. In case of error value remains unchanged
+      * @param use function to atomically modify value contained in `MVar`
+      * @return `F[A]` modified value contained in `MVar`
+      */
+    def bracketUpdate(use: A => F[A])(implicit FG: Guarantee[F], M: Monad[F]): F[A] =
+      M.flatMap(mvar.take)(oldA =>
```
Odomontois, line 46: Yeah, that should be FG.bracketOpt(mvar.take)

core/src/main/scala/tofu/syntax/bracket.scala
```diff
@@ -28,6 +28,18 @@ object bracket {
     def guaranteeAlways[B](release: F[B])(implicit F: Applicative[F], FG: Guarantee[F]): F[A] =
       FG.bracket(F.unit)(_ => fa)((_, _) => release)
 
+    def guaranteeOpt[B](use: A => F[A])(release: A => F[B])(implicit FG: Guarantee[F], A: Applicative[F]): F[A] =
```
Odomontois, line 31: No. The `guaranteeOpt` should be just `(release: Boolean => F[B])`: F[A]

core/src/main/scala/tofu/syntax/bracket.scala
```diff
@@ -28,6 +28,18 @@ object bracket {
...
+      FG.bracket(FG.bracket(fa)(use) {
+        case (oldA, success) => release(oldA).whenA(!success)
+      })(newA => release(newA).map(_ => newA)) { (_, _) => A.unit }
+
+    def guaranteeModifyOpt[B, C](
```
Odomontois, line 36: I can't understand goal of that.

core/src/main/scala/tofu/syntax/bracket.scala
```diff
@@ -12,32 +12,117 @@ import tofu.{Finally, Guarantee}
...
+      * Guarantee finalization of [[fa]].
+      * `release` is always called and boolean indicating success of usage is passed to it
+      * @param release function to release value contained in [[fa]] depending on success of usage
+      * @return `F[A]` updated value
+      */
+    def guaranteeOpt[B](release: Boolean => F[B])(implicit F: Applicative[F], FG: Guarantee[F]): F[A] =
+      FG.bracket(F.unit)(_ => fa)((_, success) => release(success))
+
+    /**
+      * Update value in [[fa]]. In case of error old value of [[fa]] is passed to `commit`, otherwise
```
Odomontois, line 64: Well, unsuccessful result is not only the error case, but could be also a cancellation.
I suppose it should be rephrased

core/src/main/scala/tofu/syntax/bracket.scala
```diff
@@ -12,32 +12,117 @@ import tofu.{Finally, Guarantee}
...
+      * result of `use` is passed.
+      * @param use function to modify value contained in [[fa]]
+      * @param commit function to commit passed value
+      * @return `F[A]` updated value
+      */
+    def bracketTransact[B](use: A => F[A])(commit: A => F[B])(implicit FG: Guarantee[F], A: Applicative[F]): F[A] =
```
Odomontois, line 70: This is nice, but I feel the name could be misleading, some people may think it has some relation to STM.
I suggest `bracketTransform` , `bracketReplace` or something like that, emphasising we are dealing with attempt to replace _old_ with _new_

mehakun: `bracketReplace` sounds good for me. Didn't think about STM...

PR #283, opened 2020-05-27
Add syntax for `Lift[F, G].lift`

catostrophe:


core/src/main/scala/tofu/syntax/lift.scala
```diff
@@ -7,6 +7,10 @@ import cats.tagless.{FunctorK, InvariantK}
 import tofu.lift.{IsoK, Lift, Unlift}
 
 object lift {
+  implicit final class LiftSyntax[F[_], A](private val fa: F[A]) extends AnyVal {
+    def lift[G[_]](implicit lift: Lift[F, G]): G[A] = lift.lift(fa)
```
Odomontois, line 11: Are we absolutely sure it doesn't not conflict with other lifts?

catostrophe: @Odomontois which other lifts do you mean? Syntax extensions in the same object?
They all require the same implicit, but are applied to different types. I can't see any conflicts here.

Odomontois: I was afraid, that in case of `Ref[Foo, A]` scala may unapply it as `F[A] = Ref[Foo, A]` but at least in scala 2.13 it isn't the case.

PR #286, opened 2020-05-29
Fix compiler warnings

Odomontois:


core/src/main/scala/tofu/Fire.scala
```diff
@@ -30,12 +30,14 @@ trait Start[F[_]] extends Fire[F] with Race[F] {
 object Start extends StartInstances[Start]
 
 trait StartInstances[TC[f[_]] >: Start[f]] {
-  final implicit def concurrentInstance[F[_]](implicit F: Concurrent[F], nonTofu: NonTofu[F]): TC[F] =
+  final implicit def concurrentInstance[F[_]](implicit F: Concurrent[F], _nonTofu: NonTofu[F]): TC[F] = {
+    _nonTofu.ImImportant()
```
catostrophe, line 34: Why don't you add `@nowarn`?

Odomontois:
> Why don't you add `@nowarn`? 

I believe it won't work on 2.12

catostrophe:
@Odomontois 
> I believe it won't work on 2.12

I suggested 4 working options

catostrophe:
> @Odomontois
> 
> > I believe it won't work on 2.12
> 
> I suggested 4 working options

`[F[_] : NonTofu]` will be the easiest one

catostrophe:
Please, fix also this:
```
[warn] /tofu/concurrent/src/test/scala/tofu/concurrent/ReadWriteTests.scala:8:26: Unused import
[warn] import cats.syntax.apply._
```

and here: /tofu/core/src/main/scala-2.12/tofu/alias/ScalaSpecificDoMonadInstances.scala
please remove `import tofu.alias.DoMonad`

PR #291, opened 2020-05-31
LocalOps alternative

oskin1:


catostrophe:
> Last time when I polished syntax for context, I tried to do the same, but found a case where inference didn't work and a working piece of code couldn't be compiled any more.
> 
> Let's try to reproduce that case.

Ahhh, CI did it for us.

oskin1:
> > Last time when I polished syntax for context, I tried to do the same, but found a case where inference didn't work and a working piece of code couldn't be compiled any more.
> > Let's try to reproduce that case.
> 
> Ahhh, CI did it for us.

@catostrophe, seems old syntax works well only when there is a single unambiguous context, but fails in other cases, for example when you have some general context in scope (`Ctx(sub: SubCtx)`) and want to work on narrower context (`fa.local((sub: SubCtx) => SubCtx(...))`)

I propose to add alternative version on `local` syntax to support cases described above

catostrophe:
> @catostrophe, seems old syntax works well only when there is a single unambiguous context, but fails in other cases, for example when you have some general context in scope (`Ctx(sub: SubCtx, ...)`) and want to work on narrower context (`fa.local((sub: SubCtx) => SubCtx(...))`)

@oskin1 I think the real reason is that it is hard to infer the type of a function `C => C` in this case:
```scala
implicit final class LocalOps[F[_], A](private val fa: F[A]) extends AnyVal {
  def local[C](project: C => C)(implicit loc: F HasLocal C): F[A] = loc.local(fa)(project)
}
```
Especially when it's passed as a lambda, the compile cannot infer it unless you provide some hints.

But when you use this syntax:
```scala
implicit final class LocalOps[F[_], A, C](private val fa: F[A])(implicit loc: F HasLocal C) {
  def local(project: C => C): F[A] = loc.local(fa)(project)
}
```
`C` type gets inferred by summoning `loc`. And yes, it must be the only `Local` in the scope.

 
> I propose to add alternative version on `local` syntax to support cases described above

I don't mind if we add extra syntax, but I definitely dislike the name `subLocal`. Why `sub`? How is it related to the problem?

oskin1:
> > @catostrophe, seems old syntax works well only when there is a single unambiguous context, but fails in other cases, for example when you have some general context in scope (`Ctx(sub: SubCtx, ...)`) and want to work on narrower context (`fa.local((sub: SubCtx) => SubCtx(...))`)
> 
> @oskin1 I think the real reason is that it is hard to infer the type of a function `C => C` in this case:
> 
> ```scala
> implicit final class LocalOps[F[_], A](private val fa: F[A]) extends AnyVal {
>   def local[C](project: C => C)(implicit loc: F HasLocal C): F[A] = loc.local(fa)(project)
> }
> ```
> 
> Especially when it's passed as a lambda, the compile cannot infer it unless you provide some hints.
> 
> But when you use this syntax:
> 
> ```scala
> implicit final class LocalOps[F[_], A, C](private val fa: F[A])(implicit loc: F HasLocal C) {
>   def local(project: C => C): F[A] = loc.local(fa)(project)
> }
> ```
> 
> `C` type gets inferred by summoning `loc`. And yes, it must be the only `Local` in the scope.
> 
> > I propose to add alternative version on `local` syntax to support cases described above
> 
> I don't mind if we add extra syntax, but I definitely dislike the name `subLocal`. Why `sub`? How is it related to the problem?

@catostrophe, I thought of `sub-` prefix in context of this syntax use-case (projecting some specific type of [sub-]context). 
But neither do I think it's best naming, so you-re free to propose better one

catostrophe:
> @catostrophe, I thought of `sub-` prefix in context of this syntax use-case (projecting some specific type of [sub-]context).
> But neither do I think it's best naming, so you-re free to propose better one

@oskin1 
I see. Sublocal is one case. But it's also a slightly better syntax extension when you pass a function with a known type, e.g. `def` rather than lambda. In this case, you don't allocate an ops object since it can be defined as AnyVal (and please do it in your PR).

I cannot find a name better than `local_`. Underscore has already been used in several cases as a marker for a very similar alternative extension.

core/src/main/scala/tofu/syntax/context.scala
```diff
@@ -37,4 +37,8 @@ object context {
   implicit final class LocalOps[F[_], A, C](private val fa: F[A])(implicit loc: F HasLocal C) {
     def local(project: C => C): F[A] = loc.local(fa)(project)
   }
+
+  implicit final class SubLocalOps[F[_], A](private val fa: F[A]) {
+    def subLocal[C](project: C => C)(implicit loc: F HasLocal C): F[A] = loc.local(fa)(project)
```
catostrophe, line 42: I propose `local_`

core/src/main/scala/tofu/syntax/context.scala
```diff
@@ -37,4 +37,8 @@ object context {
   implicit final class LocalOps[F[_], A, C](private val fa: F[A])(implicit loc: F HasLocal C) {
     def local(project: C => C): F[A] = loc.local(fa)(project)
   }
+
+  implicit final class SubLocalOps[F[_], A](private val fa: F[A]) {
```
catostrophe, line 41: extends AnyVal

catostrophe: And a better name for the ops class itself

oskin1:
@catostrophe, my only concern is that method with trailing undersocre `_` can be a bit misleading for cats users as it is usually used there for methods throwing aways thier result inside `F[_]` (such as `def parTraverse_[M[_], B](f: A => M[B])(implicit T: Foldable[T], P: Parallel[M]): M[Unit]`)

catostrophe:
> @catostrophe, my only concern is that method with trailing undersocre `_` can be a bit misleading for cats users as it is usually used there for methods throwing aways thier result inside `F[_]` (such as `def parTraverse_[M[_], B](f: A => M[B])(implicit T: Foldable[T], P: Parallel[M]): M[Unit]`)

At least this postfix looks less annoying than 0 or 1.

Issue #293, opened 2020-05-31
Logging interop modules don't get published


catostrophe:
Modules in `logging/interop` have never been released. Need to add them to a proper place in `build.sbt`.

catostrophe:
@Ssstlis attn

PR #296, opened 2020-05-31
Add logging interop modules to build.sbt properly

catostrophe:
This is to address #293

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -141,8 +141,8 @@ lazy val loggingRefined = project
   .dependsOn(loggingStr)
 
 lazy val logging = project
-  .dependsOn(loggingStr, loggingDer, loggingLayout, loggingUtil)
-  .aggregate(loggingStr, loggingDer, loggingLayout, loggingUtil)
+  .dependsOn(loggingStr, loggingDer, loggingLayout, loggingUtil, loggingShapeless, loggingRefined)
+  .aggregate(loggingStr, loggingDer, loggingLayout, loggingUtil, loggingShapeless, loggingRefined)
   .settings(defaultSettings)
 
 lazy val env = project


Ssstlis:
In favor of separating modules i think there may be is two think we can do:
- don't include interop modules into common modules like `logging`
- create another module list with interop modules, like this:
  ```scala
    lazy val interopModules = Seq(opticsInterop, zioInterop, fs2Interop, loggingRefined, loggingShapeless)
    ....
    lazy val allModuleRefs = (coreModules ++ commonModules ++ interopModules).map(x => x: ProjectReference)
    lazy val allModuleDeps = (coreModules ++ commonModules ++ interopModules).map(x => x: ClasspathDep[ProjectReference])
    ...
    lazy val tofu = project
    ...
    .aggregate((coreModules ++ commonModules ++ interopModules).map(x => x: ProjectReference): _*)
    .dependsOn(coreModules.map(x => x: ClasspathDep[ProjectReference]): _*)
  ```

Odomontois:
> In favor of separating modules i think there is maybe two thing we can do:
> 
>     * don't include interop modules into common modules like `logging`
> 
>     * create another module list with interop modules, like this:
>       ```scala
>         lazy val interopModules = Seq(opticsInterop, zioInterop, fs2Interop, loggingRefined, loggingShapeless)
>         ....
>         lazy val allModuleRefs = (coreModules ++ commonModules ++ interopModules).map(x => x: ProjectReference)
>         lazy val allModuleDeps = (coreModules ++ commonModules ++ interopModules).map(x => x: ClasspathDep[ProjectReference])
>         ...
>         lazy val tofu = project
>         ...
>         .aggregate((coreModules ++ commonModules ++ interopModules).map(x => x: ProjectReference): _*)
>         .dependsOn(coreModules.map(x => x: ClasspathDep[ProjectReference]): _*)
>       ```

I feel like this is the right way.
Lazy developer fine with all the dependencies can add just one module and get all the goodness.
Patient enterprise developer may carefully choose set of using modules for they build.
So include all the stuff by default is a good policy IMHO

PR #302, opened 2020-06-07
Tofu Streams

oskin1:
A set of abstractions & utilities for streaming effects.


streams/src/main/scala/tofu/streams/Compile.scala
```diff
@@ -0,0 +1,5 @@
+package tofu.streams
+
+trait Compile[F[_], G[_]] {
+  def compile[A](fa: F[A]): G[Seq[A]]
```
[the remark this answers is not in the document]
oskin1: @catostrophe, @FunFunFine, makes sense!
But let's focus on the hierarchy itself, maybe some type classes you think we should add in the beginning are missing?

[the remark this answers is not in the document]
catostrophe: I meant "definitely not List!"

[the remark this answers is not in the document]
oskin1: @catostrophe, ok, what then?

streams/src/main/scala/tofu/streams/Evals.scala
```diff
@@ -0,0 +1,28 @@
+package tofu.streams
+
+import cats.syntax.foldable._
+import cats.{Applicative, Foldable, Monad}
+
+trait Emits[F[_]] {
```
Odomontois, line 6: Why is this better than Alternative?

streams/src/main/scala/tofu/streams/Compile.scala
```diff
@@ -0,0 +1,5 @@
+package tofu.streams
+
+trait Compile[F[_], G[_]] {
+  def compile[A](fa: F[A]): G[List[A]]
```
Odomontois, line 4: How about accumulation of really long streams?

streams/src/main/scala/tofu/streams/Evals.scala
oskin1: @Odomontois, I can't remember any existing abstraction covering case like `def emits[A](as: List[A]): F[A]`.

catostrophe: I think Oleg implied pure+combineK

oskin1: @Odomontois, it turned out there is no lawful implementation of `Alternative` for `Stream`.

oskin1:
Example project: https://github.com/oskin1/binance-aggregator

streams/src/main/scala/tofu/streams/Evals.scala
```diff
@@ -0,0 +1,37 @@
...
+
+import cats.syntax.flatMap._
+import cats.{Applicative, Foldable, Monad, MonoidK}
+
+trait Emits[F[_]] {
+
+  val monoidK: MonoidK[F]
+  val applicative: Applicative[F]
+
+  final def emits[C[_], A](as: C[A])(implicit C: Foldable[C]): F[A] =
```
Odomontois, line 11: Do you think it's wise to make it final? So no one can't provide an optimized version

streams/src/main/scala/tofu/streams/Evals.scala
```diff
@@ -0,0 +1,37 @@
...
+
+trait Evals[F[_], G[_]] extends Emits[F] {
+
+  implicit val monad: Monad[F]
+
+  lazy val applicative: Applicative[F] = monad
+
+  def eval[A](ga: G[A]): F[A]
+
+  final def evals[C[_]: Foldable, A](gsa: G[C[A]]): F[A] =
```
Odomontois, line 32: Again. Is this really wise to have `final` here?

streams/src/main/scala/tofu/streams/Evals.scala
```diff
@@ -0,0 +1,37 @@
...
+    }
+}
+
+trait Evals[F[_], G[_]] extends Emits[F] {
+
+  implicit val monad: Monad[F]
+
+  lazy val applicative: Applicative[F] = monad
+
+  def eval[A](ga: G[A]): F[A]
```
Odomontois, line 30: Isn't this a `Lift` ?

streams/src/main/scala/tofu/streams/Evals.scala
```diff
@@ -0,0 +1,37 @@
...
+
+  final def evals[C[_]: Foldable, A](gsa: G[C[A]]): F[A] =
+    eval(gsa) >>= (emits(_))
+
+  final def evalMap[A, B](fa: F[A])(f: A => G[B]): F[B] =
```
Odomontois, line 35: Isn't this a `Lift` + `FlatMap` ?

streams/src/main/scala/tofu/streams/Merge.scala
```diff
@@ -0,0 +1,8 @@
+package tofu.streams
+
+trait Merge[F[_]] {
+
+  /** Interleaves the two inputs non-deterministically.
+    */
+  def merge[A](fa: F[A])(that: F[A]): F[A]
```
Odomontois, line 7: Feel like it should have some relation to `MonoidK`

streams/src/main/scala/tofu/streams/Pace.scala
```diff
@@ -0,0 +1,14 @@
+package tofu.streams
+
+import scala.concurrent.duration.FiniteDuration
+
+trait Pace[F[_]] {
+
+  /** Throttles `F` to the specified `rate`.
+    */
+  def metered[A](fa: F[A])(rate: FiniteDuration): F[A]
```
Odomontois, line 9: Why not just `throttle` ?

streams/src/main/scala/tofu/streams/ParFlatten.scala
```diff
@@ -0,0 +1,14 @@
+package tofu.streams
+
+trait ParFlatten[F[_]] {
```
Odomontois, line 3: Should this have some connection to `FlatMap` ?

streams/src/main/scala/tofu/streams/Region.scala
```diff
@@ -0,0 +1,16 @@
+package tofu.streams
+
+import cats.effect.ExitCase
+
+trait Region[F[_], G[_], E] {
```
Odomontois, line 5: How about parameterizing `ExitCase` as it's done in `tofu.Finally` so we can do something about zio-streams

streams/src/main/scala/tofu/streams/Broadcast.scala
```diff
@@ -0,0 +1,12 @@
+package tofu.streams
+
+trait Broadcast[F[_]] {
```
Odomontois, line 3: Is this some kind of optimization abstraction?

streams/src/main/scala/tofu/streams/Region.scala
oskin1: @Odomontois, is `Region[F[_], G[_], +Exit]` enough?

streams/src/main/scala/tofu/streams/Broadcast.scala
oskin1: @Odomontois, right, I think this functionality is too specific.

streams/src/main/scala/tofu/streams/Region.scala
Odomontois: To represent process exit you are using cats.effect.ExitCase structure, which is poorly abstracts zio.Exit or any other possible exit.
So I mean Region[F[_], G[_], Exit[_]] here

streams/src/main/scala/tofu/streams/Broadcast.scala
Odomontois: I don't mind it while you are going to support it

streams/src/main/scala/tofu/streams/Region.scala
oskin1: @Odomontois, currently I swithed to `Region[F[_], G[_], +Exit]`, which can be `Region[F[_], G[_], ExitCase[Throwable]]` in case of CatsEffect, or `Region[F[_], G[_], zio.Exit[X, E]]`. Do we really need higher-kinded type parameter `Exit[_]` here?
https://github.com/TinkoffCreditSystems/tofu/blob/1864900066f9fa1d0663cfc7954ce8f6fb17130b/streams/src/main/scala/tofu/streams/Region.scala

Odomontois: Sorry. My Bad. We can keep simple type here.

Odomontois (approved):

streams/src/main/scala/tofu/streams/Evals.scala
Odomontois: @oskin1 Actually that's interesting. What laws are broken?

fs2/src/main/scala/tofu/fs2/package.scala
```diff
@@ -0,0 +1,13 @@
+package tofu
+
+import _root_.fs2.Stream
+import tofu.lift.Lift
+
+package object fs2 {
```
catostrophe, line 6: If we now have `fs2` package, why not rename `fs2Instances` to `fs2.instances`? I never liked the name.

oskin1: @catostrophe, neither do I like `fs2Instances`, but such a renaming would break bincomp.
@Odomontois, what do you think?

catostrophe: Can we add an alias with the `@deprecated` annotation?

oskin1: @catostrophe, agree. I guess we should move `fs2Syntax` to `fs2.syntax` then as well?

oskin1: Even more, maybe we should move this syntax to tofu.streams.syntax abstracting `Stream`?

Odomontois: We may think about interop packagw like zio's

oskin1: @Odomontois, do you mean refactor `zioInterop` package in the same way?

Odomontois: Yes

Odomontois (approved):

PR #305, opened 2020-06-09
Kind enumeration description

Odomontois:
This is an internal description of enumeration of kinds to shorten names for exotic things


Odomontois:
> 
> 
> And how to deal with variances?

Right now it would create too much complexity.
We can consider variance information later and have alternative numeration for those, postfixing with Kc123 or something like that

Odomontois:
Added similar page for calculations for variance-enabled numbers

catostrophe:
Nice idea on encoding hole trees. But it's not mnemonic, not usable nor comprehendable without the codec. Is it supposed to be used internally to guarantee name uniqueness and determinism? Or did you imply using it in public APIs?

PR #309, opened 2020-06-11
Reformat

Odomontois:


env/src/main/scala/tofu/env/bio/EnvBioFunctions.scala
```diff
@@ -61,6 +61,12 @@ private[bio] trait EnvBioFunctions extends EnvBioProducts { self: EnvBio.type =>
     * }}} */
   def fromTaskTotal[A](task: Task[A]): EnvBio[Any, Nothing, A] = _ => task
 
+  def fromTaskEither[E, A](task: Task[Either[E, A]]): EnvBio[Any, Nothing, A] = _ =>
```
catostrophe, line 64: Should it return `EnvBio[Any, E, A]`?

PR #317, opened 2020-06-16
java.sql time-related datatypes loggable instances

### LINKED ISSUE #314: Provide instances of loggable for java.sql time-related classes
It would be really nice to have Loggable instances for java.sql.{Date, Timestamp, Time}.
Because not every database driver supports java.LocalDate and sometimes you just need to apply some transformations on data while logging it's content

mehakun:
fixed #314


Odomontois:
fixed #314

PR #326, opened 2020-06-23
Optics: Improved functions added monomorphic functions

Odomontois:


optics/core/src/main/scala/tofu/optics/functions/CollectionFunctions.scala
```diff
@@ -0,0 +1,56 @@
...
+import tofu.compat._
+import tofu.compat.lazySeqInstances._
+
+trait CollectionFunctions {
+  def listElemsP[A, B]: PItems[List[A], List[B], A, B] = PItems.fromTraverse[List, A, B]
+
+  def listElems[A]: Items[List[A], A] = listElemsP
+
+  def vectorElemsP[A, B]: PItems[Vector[A], Vector[B], A, B] = PItems.fromTraverse[Vector, A, B]
+  def vecElemsP[A, B]: PItems[Vector[A], Vector[B], A, B]    = PItems.fromTraverse[Vector, A, B]
```
FunFunFine, line 21: Why do we need two identical functions `vectorElemsP` and `vecElemsP`?

PR #342, opened 2020-07-23
Make optics hierarchy pic more readable

daron666:


Odomontois:
Can you also add Zipping/Grate optics which is supertype of Equivalent and subtype of Update and Upcast?

daron666:
@Odomontois signature?

daron666:
@Odomontois done. also included diagram's source file.

Odomontois:
Actually arrows are in the the wrong direction.
Zipping is supertype of Equivalent, so arrow should be from Equivalent to Zipping not otherwise.
Same for Update/Upcast with Zipping

daron666:
@Odomontois done

PR #367, opened 2020-09-03
added Unlift derivation for reader-like effects with subcontexts

klimovsv:


danslapman (dismissed):
LGTM

oskin1 (approved):

catostrophe:
@Odomontois you said "Что-то боюсь его имплиситом делать"

Odomontois:
Yeah. This function looks really good, but the amount of type parameters, that need to be inferred through the implicit search makes me think it would be rarely found.
Moreover, it could create conflicts if some ZIO implicits are reexported through context companions. That claim should be investigated though.
So while I've found this very helpful, I'd like to make this `def subContextUnlif` non-implicit at least for a while.
@klimovsv what do you think?

klimovsv:
I am okay with that. I think we should compare this derived instance with same instances for Env and ZIO. May be we could remove  particular instances in favor of this derivation.

Odomontois:
@klimovsv could you please make this def non-implicit, so I could merge now

catostrophe (approved):

PR #382, opened 2020-09-18
Refactor of Error instance hierarchy

### LINKED ISSUE #369: Inconsistency in tofu.syntax.handle
Having some `Handle[F, E]`, I can handle errors of type `E1 <: E` with the `handleWith` method, but not with `handle`: https://scastie.scala-lang.org/jzaAE02dRwuYlnD7nLaQ7Q

This is because the `handleWith` method requires `Handle[F, E1]`, which can be derived from `Handle[F, E]` using `handleDowncast`. But the `handle` method requires `HandleTo[F, F, E1]`, which can not be derived.

I think the easiest way to fix this is to require `Handle[F, E]` in the `handle` method. But I don't know if there might be cases where `HandleTo[F, F, E]` is available, but `Handle[F, E]` is not.

Odomontois:
Special thanks to @neko-kai for teaching me basic scala

fixes #369

### DIFF
--- a/core/src/main/scala/tofu/Errors.scala
+++ b/core/src/main/scala/tofu/Errors.scala
@@ -8,12 +8,18 @@ import tofu.internal.{CachedMatcher, DataEffectComp}
 import tofu.lift.Lift
 import tofu.optics.PUpcast.GenericSubtypeImpl
 import tofu.optics.{Downcast, Subset, Upcast}
+import scala.annotation.implicitNotFound
 
-trait Raise[F[_], E] extends Raise.ContravariantRaise[F, E] {
+/**
+  * Having this means a possibity of raising `E` inside type `F`
+  */
+@implicitNotFound("""can't understand how to raise ${E} inside ${F} 
+provide an instance of Raise[${F}, ${E}], cats.ApplicativeError[${F}, ${E}] or Upcast[..., ${E}]""")
+trait Raise[F[_], E] extends ErrorBase with Raise.ContravariantRaise[F, E] {
   def raise[A](err: E): F[A]
 }
 
-object Raise extends RaiseInstances with DataEffectComp[Raise] {
+object Raise extends DataEffectComp[Raise] {
 
   trait ContravariantRaise[F[_], -E] {
     def raise[A](err: E): F[A]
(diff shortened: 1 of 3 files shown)


core/src/main/scala/tofu/Errors.scala
```diff
@@ -8,12 +8,18 @@ import tofu.internal.{CachedMatcher, DataEffectComp}
 import tofu.lift.Lift
 import tofu.optics.PUpcast.GenericSubtypeImpl
 import tofu.optics.{Downcast, Subset, Upcast}
-
-trait Raise[F[_], E] extends Raise.ContravariantRaise[F, E] {
+import scala.annotation.implicitNotFound
+
+/**
+  * Having this means a possibity of raising `E` inside type `F`
```
oskin1, line 14: Why not simplifying the sentence?
```suggestion
  * Allows to raise `E` inside type `F`.
```

oskin1: Same for other similar comments below.

Odomontois: Can merge all of them, if you do multisuggestion

core/src/main/scala/tofu/Errors.scala
```diff
@@ -58,8 +65,11 @@ trait HandleTo[F[_], G[_], E] extends RestoreTo[F, G] {
     handleWith(fa)(e => G.pure(f(e)))
 }
 
-object HandleTo extends ErrorsToInstanceChain[HandleTo]
-
+/**
+  * Having this means you can recover after error of type ${E} in the ${F}
```
oskin1, line 69:
```suggestion
  * Allows to recover after an error of type ${E} in a ${F}.
```

core/src/main/scala/tofu/Errors.scala
```diff
@@ -95,46 +105,57 @@ object Handle extends HandleInstances with DataEffectComp[Handle] {
...
 }
 
-sealed class HandleInstances extends ErrorsInstanceChain[Handle] {
-  final implicit def handleDowncast[F[_], E, E1](implicit h: Handle[F, E], prism: Downcast[E, E1]): Handle[F, E1] =
-    new FromPrism[F, E, E1, Handle, Downcast] with HandlePrism[F, E, E1]
-}
-
+/**
+  * Having this means you can throw and handle errors of type ${E} in the ${F} having ${G} as recover result
+  * `G` can be same as `F` or some "subconstructor" having less errors semantically
```
oskin1, line 110:
```suggestion
  * Allows to throw and handle errors of type ${E} in a ${F} transiting to a ${G} when recovering.
  * A `G` can either be the same as `F` or some "subconstructor" having less errors semantically.
```

core/src/main/scala/tofu/Errors.scala
```diff
@@ -95,46 +105,57 @@ object Handle extends HandleInstances with DataEffectComp[Handle] {
...
-
-  final implicit def optionTIntance[F[_]](implicit F: Monad[F]): TC[OptionT[F, *], F, Unit] = new OptionTErrorsTo[F]
-
-  final implicit def eitherIntance[E]: TC[Either[E, *], Id, E] =
-    new EitherErrorsTo[E]
-  final implicit val optionTIntance: TC[Option, Id, Unit]      = OptionErrorsTo
-}
-
+/**
+  * Having this means you can throw and handle errors of type ${E} in the ${F}
```
oskin1, line 117:
```suggestion
  * Allows to throw and handle errors of type ${E} in a ${F}.
```

oskin1 (approved):

PR #383, opened 2020-09-18
Refactor of Context instance hierarchy

Odomontois:

PR #389, opened 2020-09-19
new UnliftIO instances for reader-like effects

klimovsv:


core/src/main/scala/tofu/lift/Unlift.scala
```diff
@@ -118,11 +109,32 @@ object Unlift extends UnliftInstances1 {
 }
 
 private[lift] trait UnliftInstances1 {
-  implicit def unliftReaderT[F[_]: Applicative, R]: Unlift[F, ReaderT[F, R, *]] = {
-    type RT[a] = ReaderT[F, R, a]
-    new Unlift[F, RT] {
-      def lift[A](fa: F[A]): RT[A] = ReaderT.liftF(fa)
-      def unlift: RT[RT ~> F]      = ReaderT(r => funK[RT, F](_.run(r)).pure[F])
+  implicit def unliftWithRun[F[_], G[_], A](implicit wr: WithRun[F, G, A]): Unlift[G, F] = wr
```
catostrophe, line 112: Why do we need this? WithRun extends Unlift

klimovsv: I suppose that we can summon WithRun but not Unlift. Previous instance unliftReaderT was created for same purpose as it duplicates the code from WithRun instance.

catostrophe: Probably, it should've been removed, when Unlift became a parent of WithRun. Please, make sure it's really needed.

core/src/main/scala/tofu/lift/Unlift.scala
```diff
@@ -118,11 +109,32 @@ object Unlift extends UnliftInstances1 {
...
+}
+
+private[lift] trait UnliftInstances2 {
+  implicit def unliftIOWithRun[F[_], G[_], A](implicit
```
catostrophe, line 116: Very strange chain. It's just a composition of 2 Unlifts. You require WithRun, but use it as Unlift.

catostrophe (changes requested):

core/src/main/scala/tofu/lift/Unlift.scala
Odomontois: Need to reflect this in an issue at least

core/src/main/scala/tofu/lift/Unlift.scala
klimovsv: we got _diverging implicit expansion_ if we request for 2 unlift`s, but with _WithRun_ we nicely derive UnliftIO for readers if base effect has UnliftIO or Effect. Small example https://scastie.scala-lang.org/FjgvC5IbSKitKLLAwMlK1g

core/src/main/scala/tofu/lift/Unlift.scala
klimovsv: It`s not much of a problem for me, since _unliftReaderT_ do the wrok, but we have same situation as in Provide and RunContext

catostrophe: Is this a _real_ use case when you have `ReaderT[ReaderT[IO, R1, *], R2, *]`?
You'd better use `(R1, R2)` as an environment and optics.

core/src/main/scala/tofu/lift/Unlift.scala
catostrophe: Is this a _real_ use case when you have `ReaderT[ReaderT[IO, R1, *], R2, *]`?
You'd better use `(R1, R2)` as an environment and optics.

catostrophe: I consider this instance redundant. It is too specific (IO) and uses too powerful typeclass (WithRun). It may be replaced just with Unlift composition:
`(Unlift[F, G], Unlift[G, H], FlatMap[H]) => Unlift[F, H]`

If you have diverging implicits, this instance shouldn't be implicit as many other composition instances.

catostrophe:
We need once again to review and merge #383. And only then add Unlift instance where they cannot be derived via WithRun.

core/src/main/scala/tofu/lift/Unlift.scala
klimovsv: 1. `ReaderT[ReaderT[IO, R1, *], R2, *]` is not real use case, just an example of how we can derive `UnliftIO`. In my case i have `I[_]: Effect` and locally defined `type App[A] = ReaderT[I, Ctx, A]` (or `App[_]: WithRun[*[_], I, Ctx]` in _context bounds_) and in some places i want `UnliftIO` instance for `App`. I can make it work by hand with composition, but my intention was to derive it implicitly without _diverging implicits_. I know about `Unlift[F, G].effect` and that we can create `Effect` instance for readers, but you need to write it by hand and it adds a bit boilerplate. So in cases when are happy with `F[F~>IO]`, we can get it for free. 
2. Unlift composition already exist by non-implicit `andThen` method in `Unlift`.

catostrophe: I think in your case it should work right now, without extra changes.

PR #395, opened 2020-09-26
BifunctorK and syntax.

oskin1:


catostrophe:
I don't think we should add this. It's inconsistent with the `cats-tagless` hierarchy because of constraints. But it also doesn't fit in the future hierarchy of constrained HK typeclasses, which we discussed in the tg user group.

We'd better try to design a replacement for `cats-tagless` resolving its downsides.

PR #397, opened 2020-09-28
Named composition

Odomontois:


oskin1:
@Odomontois, what is the use case of named optics?

Odomontois:
> @Odomontois, what is the use case of this?

Basically it's composition of `toString` for debugging reasons
If you are going to do some logging\error output using implicit optics sometimes it's useful to know, what is the formula for the optic, especiall if it led to some error. Since optics support point-free composition, we can sometimes print-out the whole structure of the given accessor.

oskin1 (approved):

PR #414, opened 2020-10-09
Scaladoc for Context.scala

FunFunFine:
~~It's WIP for now, but feel free to comment and improve it.~~

Note that examples do not compile and never will.

### DIFF
--- a/core/src/main/scala/tofu/Context.scala
+++ b/core/src/main/scala/tofu/Context.scala
@@ -10,7 +10,7 @@ import tofu.syntax.funk._
 trait ContextBase
 
 object ContextBase {
-  final implicit def readerTContext[F[_] : Applicative, C]: WithRun[ReaderT[F, C, *], F, C] =
+  final implicit def readerTContext[F[_]: Applicative, C]: WithRun[ReaderT[F, C, *], F, C] =
     new WithRun[ReaderT[F, C, *], F, C] {
       def lift[A](fa: F[A]): ReaderT[F, C, A] = ReaderT.liftF(fa)
 
@@ -19,10 +19,9 @@ object ContextBase {
       def local[A](fa: ReaderT[F, C, A])(project: C => C): ReaderT[F, C, A] = fa.local(project)
 
       val functor: Functor[ReaderT[F, C, *]] = Functor[ReaderT[F, C, *]]
-      val context: ReaderT[F, C, C] = ReaderT.ask[F, C]
+      val context: ReaderT[F, C, C]          = ReaderT.ask[F, C]
     }
 }
-
 
 /** Declares that [[F]] can provide value of type  Ctx
   *
@@ -35,9 +34,7 @@ object ContextBase {
   * @example
   * One common use of this is to make contextual logging:
   * {{{
-  *    import tofu.Console
   *    import tofu.syntax.console._
-  *    import tofu.HasContext
   *
   *    case class MyCtx(id: Int)
   *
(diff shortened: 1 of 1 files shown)


core/src/main/scala/tofu/Context.scala
```diff
@@ -12,48 +12,126 @@ trait ContextBase
...
+  * {{{
+  *    import tofu.syntax.console._
+  *
+  *    case class MyCtx(id: Int)
+  *
+  *    def contexualConsolling[F[_]: Console](message: String)
+  *                             (implicit hasMyCtx: F HasContext MyCtx): F[Unit] =
+  *       hasMyCtx.askF(
+  *           ctx => puts"$message (Also context: $ctx)"
+  *      )
```
oskin1, line 45: Code samples must definitely be formatted better

oskin1: Using the same formatter as for main code.

core/src/main/scala/tofu/Context.scala
```diff
@@ -80,66 +162,149 @@ object WithContext {
...
 object Provide {
   def apply[F[_]](implicit p: Provide[F]): HasProvide[F, p.Lower, p.Ctx] = p
+
   type Aux[F[_], G[_], C] = HasProvide[F, G, C]
 
   final implicit def readerTContext[F[_]: Applicative, C]: HasProvide[ReaderT[F, C, *], F, C] =
     ContextBase.readerTContext[F, C]
 }
 
+/** Synonym for [[Local]] with `C` as `Ctx` and `G` as `Lower` for better type inference
```
catostrophe, line 285: This is wrong

core/src/main/scala/tofu/Context.scala
```diff
@@ -80,66 +162,149 @@ object WithContext {
...
+
+  /** Same as [[runContext]] but higher-kinded.
+    *
+    * @example {{{
+    *   trait ProcessHandler[G[_]] {
+    *     def mapK[M[_]](fk: G ~> M): ProcessHandler[M] = ???
+    *     //...other methods
+    *   }
+    *   type WithMyContext[F[_], A] = ReaderT[F, MyCtx, A]
+    *   val processHandler: ProcessHandler[IO WithMyContext *] = ???
```
klimovsv, line 250: I think you should swap names. `ProcessHandler[IO WithContext *]` is contextual and `ProcessHandler[IO]` is not because you created it with constant context so you can't change it, providing after method calls.

core/src/main/scala/tofu/Context.scala
```diff
@@ -80,66 +162,149 @@ object WithContext {
...
+  *
+  * @tparam F context-aware effect e.g.`ReaderT[Lower, Ctx, *]`
+  */
 trait Provide[F[_]] extends ContextBase {
   type Ctx
+
+  /** Result effect type without any notion of context */
   type Lower[A]
 
+  /** Runs `fa` providing `ctx` to it.
```
klimovsv, line 233: You can reuse example from Local, but instead of creating constant context, run `logUser` via `runContext` with user context

core/src/main/scala/tofu/Context.scala
```diff
@@ -66,98 +122,260 @@ object Context {
...
+    *     def mapK[M[_]](fk: G ~> M): ProcessHandler[M] = ???
+    *     //...other methods
+    *   }
+    *   type WithMyContext[F[_], A] = ReaderT[F, MyCtx, A]
+    *   val contextualProcessHandler: ProcessHandler[IO WithMyContext *] = ???
+    *
+    *   val processHandler: ProcessHandler[IO] =
+    *     processHandler.mapK(
+    *       WithProvide[IO WithMyContext *, IO, MyCtx].runContextK
+    *     )
```
oskin1, line 253: @FunFunFine, maybe we can put simpler examples here without all these vogue modules and `???`? I think such examples better suit for an article providing some use case context along with the code.

oskin1: Same for other examples

Odomontois (approved):

PR #416, opened 2020-10-11
Add Mid documentation

mikhailchuryakov:
translate Oleg's [note](https://gist.github.com/Odomontois/18503ae41f9b283227dbee727a627468)

### DIFF
--- a/docs/mid.md
+++ b/docs/mid.md
@@ -17,7 +17,7 @@ trait MyBusinessModule[F[_]] {
   def undoBusinessThing(entity: Entity): F[Respect]
 }
 ```
-Often `F` presented like some `IO` or any transformer
+Often `F` presented like some `IO`, reader, or any transformer
 
 But signature doesn't oblige to be strict. Moreover, there is no necessity to use a functor
 
--- a/docs/mid.md
+++ b/docs/mid.md
@@ -1,5 +1,5 @@
 ---
-id: HigherKind
+id: mid
 title: Mid
 ---
 
@@ -19,9 +19,9 @@ trait MyBusinessModule[F[_]]{
 ```
 Often `F` presented like some `IO` or any transformer
 
-But signature doesn't oblige to be strict. Moreover, there is no necessity to use functor
+But signature doesn't oblige to be strict. Moreover, there is no necessity to use a functor
 
-Let's start with simple example
+Let's start with an example
 ```scala
 type Pre[F[_], A] = F[Unit]
 ``` 
@@ -35,13 +35,13 @@ Apply `MyBusinessModule[F[_]]` to `Pre[F[_], *]`
   def undoBusinessThing(entity: Entity): F[Unit]
 }
 ```
-Only the effect produced without any result. It could be logging, input validation or something like that
+Only the effect is produced without any result. It could be logging, input validation, or something like that
 
 Now consider the following type
 ```scala
 type Post[F[_], A] = A => F[Unit]
 ```
-This is contravariant type. The module takes the form
+This is a contravariant type. The module takes the form
 ```scala
 //MyBusinessModule[Post[F, *]]
 {
@@ -49,7 +49,7 @@ This is contravariant type. The module takes the form
   def undoBusinessThing(entity: Entity): Respect => F[Unit]
 }
 ```
-Such an implementation of a module can express logging or validation of a computation result.
+Such an implementation of a module can express logging or validation of a computation result
 
 Completes the next type
 ```scala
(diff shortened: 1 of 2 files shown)


docs/mid.md
```diff
@@ -0,0 +1,131 @@
...
+Consider some trait
+```scala
+trait MyBusinessModule[F[_]]{
+  def doBusinessThing(entity: Entity, info: Info): F[Value]
+  def undoBusinessThing(entity: Entity): F[Respect]
+}
+```
+Often `F` presented like some `IO` or any transformer
+
+But signature doesn't oblige to be strict. Moreover, there are no necessity to use functor
```
oskin1, line 22:
```suggestion
But signature doesn't oblige to be strict. Moreover, there is no necessity to use functor
```

docs/mid.md
```diff
@@ -0,0 +1,131 @@
...
+```
+This is contravariant type. The module takes the form
+```scala
+//MyBusinessModule[Post[F, *]]
+{
+  def doBusinessThing(entity: Entity, info: Info): Value => F[Unit]
+  def undoBusinessThing(entity: Entity): Respect => F[Unit]
+}
+```
+Having been received the result the new essence could be taken to log or, for example, send the events to kafka
```
oskin1, line 52: This sentence is a bit misleading 

```suggestion
Such an implementation of a module can express logging or validation of a computation result.
```

docs/mid.md
```diff
@@ -0,0 +1,131 @@
...
+
+Applying this to the module
+```scala
+//MyBusinessModule[Mid[F, *]]
+{
+  def doBusinessThing(entity: Entity, info: Info): F[Value] => F[Value]
+  def undoBusinessThing(entity: Entity): F[Respect] => F[Respect]
+}
+```
+It is capable of what the previous ones, but also "run" `F` multiple times or not run at all
```
oskin1, line 68:
```suggestion
`Mid` provides capabilities of both `Pre` and `Post`, but also allows to "run" `F` multiple times or not to run it at all.
```

docs/mid.md
```diff
@@ -0,0 +1,131 @@
...
+
+Such middleware can be caching, retrying or another logic, which is not implemented in infrastructure, but demand additional reflection
```
oskin1, line 70:
```suggestion
Such middleware can be caching, retrying or another logic, which is not implemented in infrastructure, but requires additional reflection
```

docs/mid.md
```diff
@@ -0,0 +1,131 @@
...
+
+## Usage
+
+It turns out enough [ApplyK](https://typelevel.org/cats-tagless/api/cats/tagless/ApplyK.html). Via
```
oskin1, line 74:
```suggestion
It turns out that [ApplyK](https://typelevel.org/cats-tagless/api/cats/tagless/ApplyK.html) is enough. Via
```

docs/mid.md
```diff
@@ -0,0 +1,131 @@
...
+```scala
+def map2K[F[_], G[_], H[_]](af: A[F], ag: A[G])(f: Tuple2K[F, G, *]~> H]): A[H]
+```
+Possible to compose main and plugin results i.e. origin implementation and addictive
```
oskin1, line 78:
```suggestion
Its makes it possible to compose the result of a main computation and the result of a plug-in computation. Hence, we can also compose main module implementation and pluggable one.
```

docs/mid.md
```diff
@@ -0,0 +1,131 @@
...
+
+`map2K` call with `F = F, G = Mid[F, *], H = F` substitute, gives `MyBusinessModule[F]` and plugin `MyBusinessModule[Mid]`
+as `af` and `ag`, only remains to implements `Tuple2K[F, G, *]~> H]` i.e. `[A] (F[A], F[A] => F[A]) => F[A]` or `(fa, f) => f(fa)`
```
oskin1, line 81: This is misleading

catostrophe:
Please check grammar with Grammarly

mikhailchuryakov:
Sorry, forgot about Grammarly

docs/mid.md
```diff
@@ -0,0 +1,131 @@
...
+## Assumption
+
+Consider some trait
+```scala
+trait MyBusinessModule[F[_]]{
+  def doBusinessThing(entity: Entity, info: Info): F[Value]
+  def undoBusinessThing(entity: Entity): F[Respect]
+}
+```
+Often `F` presented like some `IO` or any transformer
```
FunFunFine, line 20: Shouldn't we add the fact that `F` could some kind of reader too? I feel like this is not being told enough 😞

docs/mid.md
```diff
@@ -0,0 +1,131 @@
...
+
+But signature doesn't oblige to be strict. Moreover, there is no necessity to use a functor
+
+Let's start with an example
+```scala
+type Pre[F[_], A] = F[Unit]
+``` 
+Despite `Pre` has type-parameter, it doesn't put any information to the result
```
FunFunFine, line 28:
```suggestion
Despite `Pre` has type-parameter `A`, it doesn't put any information to the result
```

docs/mid.md
```diff
@@ -0,0 +1,131 @@
...
+
+Apply `MyBusinessModule[F[_]]` to `Pre[F[_], *]`
+```scala
+// MyBusinessModule[Pre[F, *]]
```
FunFunFine, line 32: Why exactly is this commented out?

docs/mid.md
mikhailchuryakov: Added

docs/mid.md
mikhailchuryakov: Fixed it

Odomontois (approved):

PR #424, opened 2020-10-20
[tofu-derivation] Narrow scope to unimplemented methods only

### LINKED ISSUE #423: Cannot derive instances for algebras with some non-final methods
```scala
@derive(representableK)
trait Foo[F[_]] {
  def bar(a: List[Int]): F[Unit]
  def widen[G[x] >: F[x]]: Foo[G] = this.asInstanceOf[Foo[G]]
}
```

The current derivation mechanism cannot derive HK typeclasses for `Foo[F]` if the `widen` method is not final. (Because `F` is not in the return type of `widen`)

catostrophe:
Closes #423


catostrophe:
Not a bug

catostrophe:
Still not sure if it's safe to keep non-final members.

PR #432, opened 2020-10-24
Interop with cats mtl

### LINKED ISSUE #431: Need interop with cats-mtl
Some libs depend on cats-mtl typeclasses. It would be great to provide a module with conversions from tofu typeclasses to cats-mtl.

Ssstlis:
Close #431


core/interop/cats-mtl/src/main/scala/tofu/core/interop/catsmtl/implicits.scala
```diff
@@ -0,0 +1,67 @@
+package tofu.core.interop.catsmtl
+
+import cats.mtl.{Ask, Local, Handle => MHandle, Raise => MRaise}
+import cats.syntax.functor._
+import cats.{Applicative, Functor}
+import tofu.{Errors, Raise, WithContext, WithLocal}
+
+object implicits {
```
catostrophe, line 8: Can we have separate `implicits` object for implicits and `instances` for explicit instances.
Names of the explicit ones should be _KPACUBbIE_.

catostrophe (changes requested):

core/interop/cats-mtl/src/main/scala/tofu/core/interop/catsmtl/implicits.scala
Ssstlis: Should explicit methods have explicit arguments as well? (except Applicative ones)

core/interop/cats-mtl/src/main/scala/tofu/core/interop/catsmtl/implicits.scala
```diff
@@ -0,0 +1,67 @@
...
+
+  implicit def deriveLocal[F[_]: Applicative, C](implicit L: WithLocal[F, C]): Local[F, C] = new Local[F, C] {
```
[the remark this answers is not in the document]
Ssstlis: My bad

core/interop/cats-mtl/src/main/scala/tofu/core/interop/catsmtl/implicits.scala
```diff
@@ -0,0 +1,67 @@
...
+    override def local[A](fa: F[A])(f: C => C): F[A] = L.local(fa)(f)
```
catostrophe, line 11: I think that the `override` word is syntax noise when you implement a trait with all abstract members. I would omit it everywhere unless it's _real_ overriding.

Ssstlis: I'm already done it locally, but now I've stuck with `diverging implicit expansion`.

core/interop/cats-mtl/src/main/scala/tofu/core/interop/catsmtl/implicits.scala
catostrophe: Good question. Let's try to think of a real example where it may be useful.

Ssstlis: I think we should place `TOFU`/`MTL` instance explicit and leave others as implicit.

core/interop/cats-mtl/src/main/scala/tofu/core/interop/catsmtl/CatsMTLTofuInstances.scala
```diff
@@ -0,0 +1,28 @@
+package tofu.core.interop.catsmtl
+
+import cats.{Applicative, Functor}
+import cats.mtl.{Ask, Local, Handle => MHandle, Raise => MRaise}
+import tofu.{Errors, Raise, WithContext, WithLocal}
+
+private[catsmtl] object CatsMTLTofuInstances {
+  private[catsmtl] class CatsMTLAskInstance[F[_], C](L: WithContext[F, C], A: Applicative[F]) extends Ask[F, C] {
+    def applicative: Applicative[F] = A
+    def ask[E2 >: C]: F[E2]         = A.map(L.context)(identity(_: E2))
```
catostrophe, line 10: just `.widen` (more efficient)
here and everywhere

catostrophe:
@Odomontois r u ok with the naming?

catostrophe (approved):

PR #450, opened 2020-11-16
Display 

FunFunFine:


core/src/main/scala/tofu/common/Display.scala
```diff
@@ -0,0 +1,19 @@
+package tofu.common
+import cats.{Show, Eval}
+
+/** Configurable and performant conversion to String */
+trait Display[A] {
+  def displayBuild(precedence: Int, cfg: DisplayConfig, a: A): Eval[Vector[String]]
```
gurinderu, line 6: Should be Eval[Chain

core/src/main/scala/tofu/common/Display.scala
```diff
@@ -0,0 +1,19 @@
...
+
+  def display(a: A): String = displayBuild(0, DisplayConfig.default, a).value.mkString
+}
+
+object Display {
+  implicit def show[A: Display]: Show[A] = (a: A) => Display[A].display(a)
+  def apply[A: Display]: Display[A] = implicitly
+}
+final case class DisplayConfig()
```
gurinderu, line 15: Should have at least lineSeparator = '\n' and indent = '\t'

core/src/main/scala/tofu/common/Display.scala
```diff
@@ -0,0 +1,19 @@
+package tofu.common
+import cats.{Show, Eval}
+
+/** Configurable and performant conversion to String */
+trait Display[A] {
```
gurinderu, line 5: extends Show[A] 

display = show

core/src/main/scala/tofu/common/Display.scala
Odomontois: I strongly disagree. 
Using standard collections here is preferred for future support, and Chain most likely has no benefits here

build.sbt
```diff
@@ -40,7 +40,7 @@ lazy val higherKindCore = project settings (
 lazy val core = project dependsOn (opticsCore, higherKindCore) settings (
   defaultSettings,
   publishName := "core",
-  libraryDependencies ++= Seq(catsCore, catsEffect, catsTagless),
+  libraryDependencies ++= Seq(catsCore, catsEffect, catsTagless, derevo, magnolia),
```
Odomontois, line 43: No, please don't add the `magnolia`and `derevo` dependencies to the `higherKindedCore`

Odomontois: I suppose this type should be in the `core` or `data` modules, and derivation in the `data-derivation` module.

core/src/main/scala/tofu/common/Display.scala
```diff
@@ -0,0 +1,101 @@
...
+  }
+
+}
+
+trait DisplayInstances {
+  def fromShow[A: Show]: Display[A]                      = (precedence: Int, cfg: Display.Config, a: A) => Eval.now(NEC.one(Show[A].show(a)))
+  implicit lazy val intDisplay: Display[Int]             = fromShow(cats.instances.int.catsStdShowForInt)
+  implicit lazy val stringDisplay: Display[String]       = (precedence: Int, cfg: Display.Config, a: String) =>
+    Eval.now(NEC.one(s""""$a""""))
+  implicit lazy val doubleDisplay: Display[Double]       = fromShow(cats.instances.double.catsStdShowForDouble)
```
Odomontois, line 85: We definitely want much more instances here

Odomontois (changes requested):

core/src/main/scala/tofu/common/Display.scala
FunFunFine: I'll stick for Vector for now. We can change it in the future though if it will be necessary due to performance issues etc.

core/src/main/scala/tofu/common/Display.scala
```diff
@@ -0,0 +1,72 @@
+package tofu.common
+import cats.{Eval, Show}
+
+/** Configurable and performant conversion to String */
+trait Display[A] extends Show[A] {
```
catostrophe, line 5: How good is the idea of extending Show?

core/src/main/scala/tofu/common/Display.scala
```diff
@@ -0,0 +1,72 @@
...
+  implicit class displayOps[A: Display](a: A) {
+    def display(config: Display.Config = Display.Config.default): String = Display[A].display(a, config)
+  }
+
+}
+trait DisplayInstances {
+  def fromShow[A: Show]: Display[A]                       = (cfg: Display.Config, a: A) => Eval.now(Vector(Show[A].show(a)))
+  implicit lazy val intDisplay: Display[Int]              = fromShow(cats.instances.int.catsStdShowForInt)
+  implicit lazy val stringDisplay: Display[String]        = (cfg: Display.Config, a: String) =>
+    Eval.now(Vector("\"" + a + "\""))
```
catostrophe, line 61: What about `"` escaping?
Why is the quote symbol `"` hardcoded?

derivation/src/test/scala/tofu/data/derived/DisplaySpec.scala
```diff
@@ -0,0 +1,146 @@
...
+    }
+
+    describe("simple cases") {
+      val bar = Bar(
+        3,
+        "abc"
+      )
+
+      val expectedBar =
+        """Bar{
```
catostrophe, line 35: I don't like this default style with no space between the class name and the `{`

core/src/main/scala/tofu/common/Display.scala
FunFunFine: This is the only way to operate with code that requires Show without using nasty imports.

Odomontois:
closing this, feel free to reopen

core/src/main/scala/tofu/common/Display.scala
FunFunFine: Which instances should we also add?

Odomontois (approved):

PR #462, opened 2020-11-26
Fix instances for layered ZIO

catostrophe:


zio/core/src/main/scala/tofu/zioInstances/ZioInstances.scala
```diff
@@ -61,7 +61,10 @@ private[zioInstances] class ZioInstances {
   final def zioTofuWithRunInstance[R, E]: ZioTofuWithRunInstance[R, E] =
     zioTofuWithRunInstanceAny.asInstanceOf[ZioTofuWithRunInstance[R, E]]
 
-  final def zioTofuUnliftHasInstance[R <: Has[_], E, C: Tag]: ZioTofuUnliftHasInstance[R, E, C] =
+  final def zioTofuUnliftHasInstance[R <: Has[_], R1 <: Has[_], E, C: Tag](implicit
```
Odomontois, line 64: That breaks at least my code with all the type arguments filled.
I wonder if I can just remove those, still, maybe it worth add another 5-ary method.

build.sbt
```diff
@@ -400,5 +400,5 @@ lazy val publishSettings = Seq(
   )
 )
 
-addCommandAlias("fmt", "all scalafmtSbt scalafmt test:scalafmt")
-addCommandAlias("checkfmt", "all scalafmtSbtCheck scalafmtCheck test:scalafmtCheck")
+addCommandAlias("fmt", "all tofu/scalafmtSbt tofu/scalafmtAll")
```
Odomontois, line 403: What's this about

catostrophe: When you switch to some module and call `fmt` you may miss the files you touched outside this module (e.g. build.sbt or dependent modules) and then your PR fails.
The new setting is for GLOBAL formatting.

zio/core/src/main/scala/tofu/zioInstances/ZioInstances.scala
catostrophe: Yes, you're right. Fixed

Odomontois (approved):

PR #473, opened 2020-12-05
Support for `WithContext` structure promotion added to `Context.Companion`

oskin1:


core/src/main/scala/tofu/Context.scala
```diff
@@ -113,16 +113,24 @@ object Context {
     * object MyContext extends Context.Companion[MyContext]
     * }}}
     */
-  trait Companion[C] {
+  trait Companion[C] extends CompanionInstances[C] {
     type Has[F[_]] = WithContext[F, C]
 
-    implicit def tofuPromoteContextStructure[F[_], A](implicit
+    implicit def promoteContextStructure[F[_], A](implicit
```
catostrophe, line 119: I think the priorities are wrong. `WithLocal` as a more concrete instance should be higher, and `WithContext` - lower.

Can you write a summon test?

core/src/main/scala/tofu/Context.scala
```diff
@@ -113,16 +113,30 @@ object Context {
     * object MyContext extends Context.Companion[MyContext]
     * }}}
     */
-  trait Companion[C] {
+  trait Companion[C] extends ContextInstances[C] {
     type Has[F[_]] = WithContext[F, C]
 
-    implicit def tofuPromoteContextStructure[F[_], A](implicit
+    abstract class WithContextOpaque[F[_], C0] extends WithContext[F, C0]
```
catostrophe, line 119: Why did you make it an inner class? It could be a concrete high-level class, as we usually do in tofu:
```scala
class WithContextContains[F[_], C, A](implicit wc: WithContext[F, C],  field: C Contains A)
  extends WithContext[F, A] {
  private val wa =  wc.extract(field) 
  def functor: Functor[F] = wa.functor
  def context: F[A]       = wa.context
}
```

And then in trait.
```scala
implicit def promoteContextStructure[F[_], A](implicit withContext: WithContext[F, C],
    field: C Contains A
): WithContextContains[F, C, A] =
    new WithContextContains[F, C, A]
```

oskin1: @catostrophe, my bad, we'd better keep style consistent.

catostrophe (approved):

Odomontois (approved):

PR #508, opened 2021-02-11
Fix package name for ConsoleContextLayout doc & bug in CCL

FunFunFine:


build.sbt
```diff
@@ -3,7 +3,7 @@ import com.typesafe.sbt.SbtGit.git
 
 moduleName := "tofu"
 
-val libVersion = "0.9.1"
+val libVersion = "0.9.3"
```
catostrophe, line 6: Why?

logging/layout/src/main/scala/tofu/logging/logback/ConsoleContextLayout.scala
```diff
@@ -31,7 +31,11 @@ class WrappedEvent(event: ILoggingEvent) extends ILoggingEvent {
       case ContextMarker(ctx, _) => intoMdc(ctx)
       case _                     =>
     }
-    for (arg <- event.getArgumentArray) intoMdc(arg)
+    val argumentArray: Array[AnyRef] = event.getArgumentArray
```
catostrophe, line 34: Simply
```scala
for (arr <- Option(event.getArgumentArray); arg <- arr) intoMdc(arg)
```

build.sbt
FunFunFine: Because it is a bugfix.
Semver tells to update minor version on bugfixes AFAIR

catostrophe: It's not you (or another contributor) who bumps the version. Please remove it and leave to those who prepare the release

FunFunFine: Got it.
```suggestion
val libVersion = "0.9.1"
```

catostrophe (dismissed):

catostrophe (approved):

PR #512, opened 2021-02-15
Draft: Logs with macro location.

FunFunFine:
This is a very rough draft of location logging.
## What
The main idea is to add implicitly summoned location to each method of the logger and to pass it right to the slf4j logger factory, which will create a logger and cache it for later use.

I can see two ways to achieve this:
- We can alter `LoggingBase` and add `Location` to its `write` method so that we will get other infrastructure around for free. There can be inconveniences though where a location could be lost.
- We can create separate `LocationLogging` just as I've done here. It requires some code repetition, but it feels safer.

Macro for Location is very-very naive of course, it works not the way we all suppose it should.

Currently, the result of the ran test looks like that:
0    [ScalaTest-run-running-LocationLoggingSpec] INFO  Users.voytsishevskiy.tofu.logging.structured.src.test.scala.tofu.logging.location.LocationLoggingSpec  - Whoooo I am locationed

## Why

With this logging thing, users will be able to create a single logger for each effect and pass it to their services, without the clunky need to make factory methods effectful or to create a special logger for each service.


logging/structured/src/main/scala/tofu/logging/location/LocationLogging.scala
```diff
@@ -0,0 +1,110 @@
...
+    def impl(c: blackbox.Context): c.Tree = {
+      import c.universe._
+      val file = c.enclosingPosition.source.path.replace("/", ".").dropRight(6).drop(1) //this is adhoc af and just for POC purpose
+      val line = c.enclosingPosition.line
+      q"_root_.tofu.logging.location.Location($file, $line)"
+    }
+  }
+}
+
+trait LocationLogging[F[_]] {
```
Odomontois, line 30: I don't think we need another implicit bound for that. Too much migration pain for people who just want to add some functionality to their existing logs

logging/structured/src/main/scala/tofu/logging/location/LocationLogging.scala
```diff
@@ -0,0 +1,110 @@
...
+
+  def withContext[I[_]: Sync, F[_]: Sync](implicit ctx: LoggableContext[F]): I[LocationLogging[F]] = {
+    import ctx.loggable
+
+    Sync[I].delay(LoggerFactory.getILoggerFactory).map { factory =>
+      new LocationLogging[F] {
+        override def writeLocation(level: Logging.Level, location: Location, message: String, values: LoggedValue*)
+            : F[Unit] =
+          for {
+            logger <- Sync[F].delay(factory.getLogger(location.loggerName)) //cached inside factory
```
Odomontois, line 103: This comment is excessive and misleaading

Odomontois: Still I dont understand the whole point of `I` outside if we acquire the logger every time

Odomontois (changes requested):

logging/structured/src/main/scala/tofu/logging/Logging.scala
```diff
@@ -19,50 +20,68 @@ import scala.reflect.ClassTag
...
+  /** Push new message to log, level will be automatically checked
+    *
     * @param level desired level of logging, message will not be rendered\sent, if logging level
     *              of current logger is not low enough
     * @param message composed string with `{}` placeholders for values
     *                 do not create strings on each call, use constant template string instead
     * @param values  log parameters , values of types having `Loggable` instance would be converted automatically
     */
-  def write(level: Level, message: String, values: LoggedValue*): F[Unit]
+  def write(level: Level, location: Option[Location], message: String, values: LoggedValue*): F[Unit]
```
Odomontois, line 31: This change is not binary compatible.
It would be better if old methods kept the same

Odomontois (changes requested):

Odomontois:
closing this, feel free to reopen

PR #537, opened 2021-03-02
Enable newtype derivation for loggable

Odomontois:


logging/derivation/src/main/scala/tofu/logging/derivation/loggable.scala
```diff
@@ -4,8 +4,9 @@ package derivation
 import cats.Show
 import magnolia.{CaseClass, Magnolia, SealedTrait}
 import derevo.Derivation
+import derevo.NewTypeDerivation
 
-object loggable extends Derivation[Loggable] {
+object loggable extends Derivation[Loggable] with NewTypeDerivation[Loggable] {
```
FunFunFine, line 9: the same needs to be done to show derivation in `tofu.logging.derivation.show`

catostrophe (changes requested):
fix what Anton says

catostrophe (approved):

Issue #292, opened 2020-05-31
Consider changing scalafmt rules


catostrophe:
I do support the need for a forced formatter in a project with many contributors, but the current ruleset is awful and doesn't work properly in IDEA (has some wrong fields).

catostrophe:
@ReiReiRei attn

ReiReiRei:
@catostrophe do not mind. Rules had been set by @Odomontois so I have decided do not change them. Anyway, if rules are changed it meens a new large commit with reformated code. Anyway I think @catostrophe  should create new issue for IDEA for better scalafmt support.

catostrophe:
I tried [.scalafmt.conf](https://github.com/7mind/izumi/blob/develop/.scalafmt.conf) from Izumi repo.
Liked the result, but I would suggest a few changes:
* `maxColumn = 120`
* drop `newlines.alwaysBeforeCurlyBraceLambdaParams = true`

PR #567, opened 2021-03-19
Factorize and cooler Logs

Odomontois:


higherKindCore/src/test/scala/tofu/higherKind/SyntaxCheck.scala
```diff
@@ -0,0 +1,5 @@
+package tofu.higherKind
+
+object SyntaxCheck {
+//   def foo[F[_, _]: MonoidBK]
```
catostrophe, line 4: What's this?

logging/structured/src/main/scala/tofu/logging/LoggableInstances.scala
```diff
@@ -0,0 +1,192 @@
...
+    ): R                                   =
+      receiver.addFloat(name, a, input)
+  }
+
+  final implicit val booleanLoggable: Loggable[Boolean] = new SingleValueLoggable[Boolean] {
+    def logValue(a: Boolean): LogParamValue = BoolValue(a)
+    override def putField[I, V, R, M](a: Boolean, name: String, input: I)(implicit
+        receiver: LogRenderer[I, V, R, M]
+    ): R                                    =
+      receiver.addBool(name, a, input)
```
oskin1, line 97: This formatting hurts my eyes. Let's do smt about it.

Odomontois: there is issue about it https://github.com/tofu-tf/tofu/issues/292

higherKindCore/src/test/scala/tofu/higherKind/SyntaxCheck.scala
Odomontois: some garbare, removed it

logging/derivation/src/main/scala/tofu/logging/derivation/loggingMid.scala
```diff
@@ -0,0 +1,31 @@
...
+
+import tofu.higherKind.derived.HigherKindedMacros
+import derevo.DerivationKN3
+import derevo.DerivationKN11
+import tofu.logging.bi.LoggingBiMid
+import tofu.logging.bi.LoggingBiMidBuilder
+import derevo.PassTypeArgs
+import derevo.ParamRequire
+
+object loggingMid
```
FunFunFine, line 12: We definitely need some scaladoc here, probably with the link to the docs.
Those are things that have to be described:
- differences between all of `loggingMid*`
- how each method will be logged
- how to configure logging
- where the `Logging` instance will be looked for

logging/structured/src/main/scala/tofu/logging/LoggingMid.scala
```diff
@@ -0,0 +1,157 @@
...
+
+  def prepare[Alg[_[_]]](implicit Alg: ClassTag[Alg[Any]]): Prepared[Alg] = new Prepared[Alg](Alg.runtimeClass)
+
+  class Method[U[f[_]], Res: Loggable](
+      cls: Class[_],
+      method: String,
+      args: mutable.Buffer[(String, LoggedValue)]
+  ) {
+    def arg[A: Loggable](name: String, a: A): this.type = {
+      args += (name -> a)
```
FunFunFine, line 53: Shouldn't this code be marked as unsafe somehow with comments or something? 
The reader would need to understand the intentions behind the decision to make it mutable and such.

logging/structured/src/main/scala/tofu/logging/LoggingMid.scala
```diff
@@ -0,0 +1,157 @@
...
+
+object LoggingMid extends LoggingMidBuilder.DefaultImpl {
+  type Result[A] = LoggingMid[A]
+  def instance[U[_[_]]]: U[LoggingMid] = macro HigherKindedMacros.factorizeThis[U]
+
+  type Of[U[_[_]]] = U[LoggingMid]
+}
+
+/** Logging middleware generator */
+trait LoggingMidBuilder {
```
FunFunFine, line 32: Somebody once told me that mixing up "utility" code like all those bulders and mutable stuff together with actual API  hurts readability and discoverability. 
One can almost sink in all this stuff.
Is it possible to hide it in some folders or in other files to keep things clean?

WDYT?

logging/structured/src/main/scala/tofu/logging/impl/UniversalLogging.scala
```diff
@@ -0,0 +1,81 @@
...
+      logger: Logger,
+      marker: Marker,
+      message: String,
+      values: Seq[LoggedValue]
+  ): Unit                                                                                                =
+    level match {
+      case Trace => logger.trace(marker, message, values: _*)
+      case Debug => logger.debug(marker, message, values: _*)
+      case Info  => logger.info(marker, message, values: _*)
+      case Warn  => logger.warn(marker, message, values: _*)
```
FunFunFine, line 36: Those are impure as far as I see, so they should be marked as such somehow or even made private. I am like 99% sure that none of the tofu users will want to use the unsafe logger.

FunFunFine (approved):
V E L I C H I E is as close as it never was.

One side note: I think the segregation between unsafe impure code has to be more clear. Tofu users should easily see that some code doesn't break RT or, if they need to break it, it should be clear too. Something like `import tofu.logging.unsafe`.
When there's no such difference it can lead to false assumptions or annoying bugs.
Although this can be discussed in the background and be done later.

higherKindCore/src/main/scala/tofu/higherKind/bi/MonoidBK.scala
```diff
@@ -0,0 +1,25 @@
...
+    def combineK[A](x: F[A, X], y: F[A, X]): F[A, X] = combinebk(x, y)
+
+    def empty[A]: F[A, X] = emptybk
+  }
+
+  def monoid[X, Y]: Monoid[F[X, Y]] = new Monoid[F[X, Y]] {
+    def combine(x: F[X, Y], y: F[X, Y]): F[X, Y] = combinebk(x, y)
+
+    def empty: F[X, Y] = emptybk
+  }
```
catostrophe, line 24: These ones might be implemented as vals + asInstanceOf, although it's not worth extra efforts.

higherKindCore/src/main/scala/tofu/higherKind/bi/SemigroupBK.scala
```diff
@@ -0,0 +1,18 @@
...
+
+  def semigroupK[X]: SemigroupK[F[X, *]] = new SemigroupK[F[X, *]] {
+    def combineK[A](x: F[X, A], y: F[X, A]): F[X, A] = combinebk(x, y)
+  }
+
+  def leftSemigroupK[X]: SemigroupK[F[*, X]] = new SemigroupK[F[*, X]] {
+    def combineK[A](x: F[A, X], y: F[A, X]): F[A, X] = combinebk(x, y)
+  }
+
+  def semigroup[X, Y]: Semigroup[F[X, Y]] = combinebk(_, _)
```
catostrophe, line 17: same here

logging/structured/src/main/scala/tofu/logging/LoggingMid.scala
Odomontois: This is not a utility code.
Currently I have a habit of hiding utility things in the `impl` package.
But this one is something like a configuration as a code interface, one can easily craft its own logging mid derivation using this builder.

logging/structured/src/main/scala/tofu/logging/LoggingMid.scala
Odomontois: I suppose we can split this unsafe behaviour into the package private implementation of the abstract class, as OOP grandfather behest says

FunFunFine (approved):

PR #669, opened 2021-06-07
Fix Tinkoff link

kyusu:
The link in the user showcases currently links to https://docs.tofu.tf/en/tinkoff.ru because it is only relative.


oskin1:
Why do we need Tinkoff there at all?

Odomontois:
> Why do we need Tinkoff there at all?

We need all users that want to appear at docs page. 
This is the users section, provided by the template

PR #680, opened 2021-07-01
Timeout moved to kernel

Odomontois:


catostrophe (dismissed):

kernel/src/main/scala/tofu/Timeout.scala
```diff
@@ -0,0 +1,16 @@
+package tofu
+
+import scala.concurrent.duration.FiniteDuration
+import tofu.internal.EffectComp
+import tofu.internal.Interop
+
+trait Timeout[F[_]] {
+  def timeoutTo[A](fa: F[A], after: FiniteDuration, fallback: F[A]): F[A]
```
catostrophe, line 8: Fallback should be `=> F[A]` but maybe not in this PR.

catostrophe (approved):

PR #681, opened 2021-07-01
Guarantee kernelize

Odomontois:
This one is tricky. 
We had to use whitebox macros to allow optional instance to derive type that may be unconstructable in the using code and shapeless Aux encoding for intermediate carrier .

### DIFF
--- a/kernel/src/main/scala/tofu/kernel/types.scala
+++ b/kernel/src/main/scala/tofu/kernel/types.scala
@@ -1,10 +1,12 @@
-package tofu.kernel
+package tofu
+package kernel
 
-import tofu.{WithContext, Local, Context}
-import tofu.WithProvide
-import tofu.WithRun
+import cats.ApplicativeError
+import cats.MonadError
 
-object types {
+object types extends KernelTypes
+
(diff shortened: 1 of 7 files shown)


kernel/src/main/scala/tofu/kernel/types.scala
```diff
@@ -1,10 +1,12 @@
...
-import tofu.{WithContext, Local, Context}
-import tofu.WithProvide
-import tofu.WithRun
+import cats.ApplicativeError
+import cats.MonadError
 
-object types {
+object types extends KernelTypes
+
+trait KernelTypes extends Any {
```
FunFunFine, line 9: Why is it here?

FunFunFine: I mean `extends Any` part

FunFunFine (approved):

kernel/src/main/scala/tofu/kernel/types.scala
catostrophe: universal trait

PR #682, opened 2021-07-01
Kernelize Start\Fire

Odomontois:
Using basically the same approach as in #681, while parameterizing Start over Fiber into new type class Fibers


kernel/src/main/scala/tofu/Fire.scala
```diff
@@ -0,0 +1,31 @@
...
+import scala.util.Either
+import tofu.internal.{EffectComp, Effect2Comp}
+import tofu.internal.carriers.FibersCarrier
+
+trait Fire[F[_]] {
+  def fireAndForget[A](fa: F[A]): F[Unit]
+}
+
+object Fire extends EffectComp[Fire] {
+  final implicit def byCarrier[F[_], Fib[_]](implicit
```
FunFunFine, line 12: I think it's better to note somehow that it's just for scala 2.12

kernel/src/main/scala/tofu/Fire.scala
```diff
@@ -0,0 +1,31 @@
...
+trait Race[F[_]] extends Fire[F] {
+  def race[A, B](fa: F[A], fb: F[B]): F[Either[A, B]]
+  def never[A]: F[A]
+}
+
+object Race extends EffectComp[Race] {
+  def never[F[_], A](implicit race: Race[F]): F[A] = race.never
+}
+
+trait Fibers[F[_], Fib[_]] extends Race[F] {
```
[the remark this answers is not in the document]
Odomontois: I tend not to mix type parameter names with existing types. It creates readability hell sometimes

FunFunFine (dismissed):

kernel/src/main/scala/tofu/Fire.scala
Odomontois: It's not

Odomontois: As in #681 this carrier allows `Fib` to be inferred within the macro, just like shapeless whitebox macros do the typelevel calculations we can depend on types that can be unknown if the needed module isn't in the classpath

kernel/src/main/scala/tofu/Fire.scala
```diff
@@ -0,0 +1,31 @@
...
+trait Race[F[_]] extends Fire[F] {
+  def race[A, B](fa: F[A], fb: F[B]): F[Either[A, B]]
+  def never[A]: F[A]
+}
+
+object Race extends EffectComp[Race] {
+  def never[F[_], A](implicit race: Race[F]): F[A] = race.never
+}
+
+trait Fibers[F[_], Exit[_], Fib[_]] extends Race[F] {
```
vilunov, line 26: Why do we need the `Exit` param? Does not seem to be used anywhere

Odomontois: - [ ] Actually use Exit

catostrophe (approved):

PR #687, opened 2021-07-05
Context typeclasses

FunFunFine:
This PR moves WithContext, WithLocal, WithProvide, and WithRun from `Context.scala`.
Also, it marks HasContext and others as deprecated.

I was trying to keep this PR as simple as possible so it's kinda small.

### DIFF
--- a/kernel/src/main/scala/tofu/Context.scala
+++ b/kernel/src/main/scala/tofu/Context.scala
@@ -8,6 +8,7 @@ import tofu.syntax.funk._
 import cats.Monad
 import tofu.kernel.types._
 import cats.arrow.FunctionK
+import tofu.internal.ContextBase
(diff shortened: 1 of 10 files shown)


Odomontois (dismissed):

kernel/src/main/scala/tofu/Context.scala
```diff
@@ -66,6 +68,7 @@ trait Context[F[_]] extends ContextBase {
 }
 
 /** Companion object for [[Context]] */
+@deprecated("Migrate to With* typeclasses", "0.10.3")
```
catostrophe, line 71: Why 0.10.3? I think it will be 0.11.0

Odomontois: It's the correct "since" version, afaic

PR #688, opened 2021-07-05
move members into With* types;

FunFunFine:
This PR prepares tofu for the removal of `HasContext` and others.

Methods and scaladocs are copied from HasContext to WithContext etc.
Other members in HasContext now proxy to the members in WithContext.

Usage of Has*-things are minimized, but without breaking any external APIs (I hope so)


catostrophe:
Can we make `WithXXX` aliases and rename the real traits to Contex/Local/etc?

FunFunFine:
> Can we make `WithXXX` aliases and rename the real traits to Contex/Local/etc?

It's a good idea, though I'm afraid it is possible only after removal of the old Context/Local/etc with abstract type members.

Odomontois (approved):

PR #693, opened 2021-07-09
[Config] Allow simple types to be parsed from strings

Odomontois:
Which is a common case when config expanded \ modified by the environment


danslapman (approved):
LGTM

PR #698, opened 2021-07-12
Continuational representation of ConnectionIO

catostrophe:
This is a CE2-compatible change. I managed to keep the main API untouched and the usage example hasn't almost changed.

## Motivation
In CE2 `ConnectionIO` used to have a `LiftIO` instance, which we used for lifting arbitrary effects into it in order to enrich the transactional logic with non-transactional actions (e.g. tracing). This approach worked quite well when the effect used in the application was simple enough like `IO` in CE2 used to be, but lifting something more complex like `ZIO`, `Env` or even `ReaderT` required turning the effect into `IO` and introducing extra means for keeping and passing the context around. See implicit derivations for lifting into `ConnectionRIO`.

Migration to CE3 and the related changes in `doobie` itself made me look at the problem from a different angle. First, we don't have a simple "run-whenever-you-want" `IO` anymore - there's no easy (or even proper) way to turn an arbitrary effect into a runnable action. Second, `ConnectionIO` can now lift only `Future`s. These changes aren't bad. Oppositely, they helped me discover another - a more universal - way of representing database effects via a continuational form of `ConnectionIO`.

Instead of lifting contextual (or even simple) effects into something like `ConnectionIO` or `ReaderT[ConnectionIO, R, *]` let's work with:
```scala
type DB[x] = ConnectionIO ~> F => F[x]  // actually I use Kleisli for better implicit derivations
```
where `F[_]` is the target effect `ConnectionIO` will be interpreted into. Its form allows us easily lift both database logic (expressed in `ConnectionIO`) and any arbitrary non-transactional actions (expressed in `F`) to the same effect without transforming them to `IO` or `Future`.

The only new thing we need is a proper transactor-interpreter where the transaction strategy and interpretation are separated. And we can create one, thanks to the `doobie` API, and get a proper `DB ~> F` transformation.

Compared to the new implementation, the old one seems much more complex and redundant. I deprecated it and the migration seems very easy (see the example).

## Further work on `LogHandler`
I haven't touched it. But there are some thoughts. Currently `EmbeddableLogHandler` relies on `UnliftIO`. This can be changed for CE3 and implemented via `Dispatcher`. So the API won't change (not sure about `async`), but the bounds on `F[_]` will.


gurinderu (approved):

oskin1 (dismissed):

Z1kkurat (dismissed):

oskin1 (approved):

modules/doobie/src/test/scala/tofu/doobie/example/TofuDoobieExample.scala
```diff
@@ -34,9 +33,11 @@ final case class Ctx(traceId: String)
 trait Logging[F[_]] {
   def info(msg: String): F[Unit]
 }
-object Logging      {
-  def make[F[_]: FlatMap: Console: HasContext[*[_], Ctx]]: Logging[F] =
+
+object Logging {
```
FunFunFine, line 37: As was told in tofu-chat (or ponv, I don't remember exactly where), this is example feels overloaded and hides the demonstrated techniques.
Is it possible to split it up a bit, like move Logging and Tracing into separate file `infrastructure.scala`? That would make it easier to read and such.

modules/doobie/src/test/scala/tofu/doobie/example/TofuDoobieExample.scala
```diff
@@ -60,6 +63,7 @@ object Tracing      {
 
 // Model
```
FunFunFine, line 64: This would fit in `model.scala` for the matter of readability.

FunFunFine (approved):
Comments regarding examples are better to be fixed later.

Odomontois (approved):

PR #701, opened 2021-07-13
Context removal prepare

FunFunFine:


Odomontois:
I suppose this is branched from the really old commit, it would be cool if the PR wouldn't touch much aside contextual sources

PR #711, opened 2021-07-16
present Time & Clock in kernel

Odomontois:
Also based on #709


modules/kernel/src/main/scala/tofu/Clock.scala
```diff
@@ -0,0 +1,67 @@
...
+
+trait Clock[F[_]] {
+
+  /** Returns the current time, as a Unix timestamp (number of time units
+    * since the Unix epoch), suspended in `F[_]`.
+    *
+    * This is the pure equivalent of Java's `System.currentTimeMillis`,
+    * or of `CLOCK_REALTIME` from Linux's `clock_gettime()`.
+    */
+  def realTime(unit: TimeUnit): F[Long]
```
FunFunFine, line 17: If tofu will have its own Clock typeclass, so why should it be exactly like cats one? Is it possible to fix some of the design problems e.g. not having information on units of measurement in the signature (ce3 style with FiniteDuration for example (but rather something better))?

modules/kernel/src/main/scala/tofu/Clock.scala
```diff
@@ -0,0 +1,67 @@
...
+    *
+    * This is the pure equivalent of Java's `System.nanoTime`,
+    * or of `CLOCK_MONOTONIC` from Linux's `clock_gettime()`.
+    */
+  def nanos: F[Long]
+}
+
+object Clock extends EffectComp[Clock] with ClockInterop
+
+trait Sleep[F[_]] {
```
FunFunFine, line 30: This one is more like Fiber-concurrency-related stuff, so I think it should have its own file

modules/kernel/src/main/scala/tofu/Clock.scala
```diff
@@ -0,0 +1,67 @@
...
+
+  /** Pauses execution for desired duration
+    */
+  def sleep(duration: FiniteDuration): F[Unit]
+}
+
+object Sleep extends EffectComp[Sleep] with SleepInterop
+
+trait Timeout[F[_]] {
```
FunFunFine, line 39: Same as with Sleep

modules/kernel/src/main/scala/tofu/concurrent/Atom.scala
```diff
@@ -58,25 +55,6 @@ object Atom {
     }
   }
 
-  final case class AtomByRef[F[_], A](ref: Ref[F, A]) extends Atom[F, A] {
```
FunFunFine, line 61: These changes were in #709, so I guess there is something wrong with branches 🤔

FunFunFine (approved):

modules/kernel/src/main/scala/tofu/Clock.scala
Odomontois: You have some specific design in mind?

modules/kernel/src/main/scala/tofu/Clock.scala
Odomontois: No, it's unrelated to Fiber and concurrency, it doesn't imply some branching of execution, so I disagree here

modules/kernel/src/main/scala/tofu/Clock.scala
Odomontois: Same answer as with Sleep

modules/kernel/src/main/scala/tofu/concurrent/Atom.scala
Odomontois: These two are just following up PRs

catostrophe (approved):

PR #716, opened 2021-07-20
Update fs2Instances package to ce3

FunFunFine:
This PR updates fs2 version and so that updates ce3 version.
I expect some other PR to do version update the right way, while this one focuses solely on the update of fs2Instances package.

Also this PR should target another branch specific for ce3-migration. I'll edit this when this branch is created.

### DIFF
--- a/modules/fs2/src/main/scala/tofu/fs2/fs2Instances.scala
+++ b/modules/fs2/src/main/scala/tofu/fs2/fs2Instances.scala
@@ -1,9 +1,11 @@
-package tofu
-package fs2Instances
+package tofu.fs2
+
 import _root_.fs2._
-import cats.effect.{Concurrent, ExitCase, Sync, Timer}
+import cats.effect.kernel.Resource
+import cats.effect.{Concurrent, Sync, Temporal => ceTemporal}
 import cats.tagless.FunctorK
 import cats.{FlatMap, Functor, ~>}
+import tofu._
 import tofu.higherKind.Embed
 import tofu.lift.Lift
 import tofu.streams._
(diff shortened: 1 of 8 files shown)


project/Dependencies.scala
```diff
@@ -16,7 +16,7 @@ object Dependencies {
 
     val cats = "2.6.1"
 
-    val catsEffect = "2.5.1"
```
Odomontois, line 19: lol

Odomontois: No, that's wont work. You can wait for #720 where both ce dependencies will be in the build

project/Dependencies.scala
```diff
@@ -16,7 +16,7 @@ object Dependencies {
...
+    val catsEffect = "3.1.1"
```
Odomontois, line 19: No, that's wont work. You can wait for #720 where both ce dependencies will be in the build

FunFunFine: Quote from the description:
> I expect some other PR to do version update the right way

modules/fs2/src/main/scala/tofu/fs2/fs2Instances.scala
```diff
@@ -62,38 +64,38 @@ private[fs2Instances] trait Fs2Instances1 extends Fs2Instances2 {
         ffa.parJoin(maxConcurrent)
     }
 
-  implicit def fs2PaceInstance[F[_]: Timer]: Pace[Stream[F, *]] =
+  implicit def fs2PaceInstance[F[_]: ceTemporal]: Pace[Stream[F, *]] =
```
Odomontois, line 67: using uncapitalized names for typeclasses is unusual
`TemporalCE`, mybe?

Odomontois (approved):

Odomontois:
Close this for now

PR #720, opened 2021-07-20
rest of the stuff

Odomontois:


modules/kernel/src/main/scala/tofu/Fire.scala
```diff
@@ -2,15 +2,23 @@ package tofu
...
+import scala.annotation.unused
 
 trait Fire[F[_]] {
   def fireAndForget[A](fa: F[A]): F[Unit]
 }
 
 object Fire extends EffectComp[Fire] {
-  final implicit def byCarrier[F[_], Ex[_], Fib[_]](implicit
-      carrier: FibersCarrier.Aux[F, Ex, Fib]
+  final implicit def byCarrier2[F[_], Ex[_], Fib[_]](implicit
```
FunFunFine, line 15: Could you use postfix `CE2` & `CE3`? I think it makes the intention clearer for a reader.
Like so `byCarrierCE2`

modules/kernel/src/main/scala/tofu/Scoped.scala
```diff
@@ -92,11 +94,11 @@ trait ScopedInstances {
   final def makeExecute[Tag, F[_]](p1: ExecutionContext): ScopedExecute[Tag, F] =
     macro Interop.delegate1p1[Execute[F], Tag, F, { val `tofu.interop.CE2Kernel.makeExecute`: Unit }]
 
-  final implicit def asyncExecute[F[_]]: Execute[F] =
-    macro Interop.delegate[Execute[F], F, { val `tofu.interop.CE2Kernel.asyncExecute`: Unit }]
+  final def makeExecute3[Tag, F[_]](p1: ExecutionContext): ScopedExecute[Tag, F] =
```
catostrophe, line 97: Inconsistent naming with `makeExecute`

Odomontois: I wonder if we need to rename existing for some amount of versions method `makeExecute` for cats-effect 2

modules/kernel/src/main/scala/tofu/Fire.scala
Odomontois: okaaay

Odomontois:
So currently we need merge of #722 to fix QVar for CE3 implementation
And copy of all the tests in core to core3

catostrophe (approved):

PR #723, opened 2021-07-22
Add collections and parallel syntax

Odomontois:


modules/higherKindCore/src/main/scala/tofu/syntax/monadic.scala
```diff
@@ -51,6 +52,10 @@ object monadic extends TupleSemigroupalSyntax with ApplicativeSyntax with MonadS
     def map2[B, Z](fb: F[B])(f: (C, B) => Z)(implicit F: Apply[F]): F[Z]                           = F.map2(fa, fb)(f)
     def map2Eval[B, Z](fb: cats.Eval[F[B]])(f: (C, B) => Z)(implicit F: Apply[F]): cats.Eval[F[Z]] =
       F.map2Eval(fa, fb)(f)
+
+    def replicate_(count: Long)(implicit F: Applicative[F], FD: Defer[F]): F[Unit] =
+      if (count <= 0) unit[F]
+      else productR(FD.defer(replicate_(count - 1)))
```
FunFunFine, line 58: `fa.productR(FD.defer(...))` reads much more easily IMHO even though it is one allocation overhead.

modules/kernel/src/main/scala/tofu/syntax/collections.scala
```diff
@@ -0,0 +1,86 @@
...
+
+import cats.syntax._
+import cats.{Applicative, Traverse, TraverseFilter}
+import cats.syntax.functor._
+import cats.data.State
+import cats.data.StateT
+import cats.Monad
+import cats.free.Free
+import cats.FlatMap
+object collections
```
FunFunFine, line 11: no newline

modules/higherKindCore/src/main/scala/tofu/syntax/monadic.scala
Odomontois: No, there is no allocation overhead, still I'm not agree it's more readable

modules/kernel/src/main/scala/tofu/syntax/collections.scala
Odomontois: I suppose it's something scalafmt should watch.
Actually we enabled scalafmt check to avoid any style discussions in reviews, so I purposedly ignore this one

FunFunFine (approved):

PR #731, opened 2021-07-29
DRAFT: better documentation for logging

FunFunFine:


docs/logging/syntax.md
```diff
@@ -0,0 +1,20 @@
...
+# tofu.logging.syntax
+
+All the usual methods are implemented with string interpolator:
+
+There are basic methods:
+- `info"This is INFO message`
+- `warn"This is WARN message`
+- `debug"This is DEBUG message`
+- `trace"This is TRACE message`
+- `trace"This is TRACE message`
```
catostrophe, line 16:
```suggestion
```

docs/logging/home-page.md
```diff
@@ -0,0 +1,52 @@
...
+
+## How to use
+
+The usage of logging is described as a set of recipes, suitable for many cases.
+Check it out [here](recipes/recipes.md).
+
+## Quick demo
+
+```scala
+@derive(loggable)
```
mehakun, line 33: I think imports need to be added since most people (according to my humble opinion) will not go further and just use derivation and syntax

docs/logging/main-entities.md
```diff
@@ -0,0 +1,80 @@
...
+
+# tofu.logging core concepts
+
+`tofu.logging` consists of three main things:
+
+## Typeclass `Loggable[A]`
+This is a type class that describes how an arbitrary instance of type A should be presented in the log messages.
+
+## Derivation
+Loggable instances could be derived automatically using `derevo`:
```
mehakun, line 15: I think it's better to put link to derevo's repo: https://github.com/tofu-tf/derevo

docs/logging/main-entities.md
```diff
@@ -0,0 +1,80 @@
...
+case class Payment(id: Long, @masked(MaskMode.Erase) cardNumber: String, @unembed name: ClientData)
+```
+
+A message `info"This is $payment"` would look like that:
+```json
+{
+  "message": "This is Payment",
+  "payment.id": 434324,
+  "to": "do"
+}
```
mehakun, line 45: cardNumber and name's data are missing

docs/logging/recipes/context.md
```diff
@@ -0,0 +1,100 @@
...
+
+## Logs creation
+
+Only the `Logs` factory carries information about context, while each of the logging instances doesn't have that anywhere in API.
+Let's define the effect the logging would happen in:
+```scala
+import cats.effect.IO
+import cats.data.ReaderT
+
+type RequestIO = ReaderT[IO, RequestContext]
```
patseev, line 41: type RequestIO[A] = ReaderT[IO, RequestContext, A]

docs/logging/home-page.md
```diff
@@ -0,0 +1,55 @@
...
+
+## What's next
+- Discover [the key features](./key-features.md)
+- Get to know [the core concepts](./main-entities.md)
+- Learn how to use [the syntax](./syntax.md)
+- Find a way to use `logging` suitable for you in the [recipes](recipes/recipes.md)
+- Check out [the examples](https://github.com/tofu-tf/tofu/tree/better-doobie-example/examples)
+
+## Old documentation
+You can also visit obsolete documentation [here](../logging.md).
```
patseev, line 55: Is there any point in leaving reference to this?
Do you ever read "obsolete documentation"?

docs/logging/key-features.md
```diff
@@ -0,0 +1,96 @@
...
+```scala
+import cats._
+import cats.effect._
+
+case class TraceId(id: Long)
+
+type TracedIO[A] = ReaderT[TraceId, IO, A]
+```
+
+tofu.logging can extract it and automatically add into the structure of every log message:
```
patseev, line 65: Line states that everything happens "automatically" when in fact you need to define a specific logger and have `loggable` for your context (which is not defined even in this example). I think it would be better to rephrase this as:

"Given that there's a defined loggable instance for this context and provided logger is context aware, every logged message will contain information about context"

Also I think it makes sense to leave a link to `context.md` here.

docs/logging/recipes/auto.md
```diff
@@ -0,0 +1,46 @@
...
+```
+
+`tofu.logging` allows to automatically generate those lines about entering and exiting the method.
+
+## Autogenerated logging
+
+You can automatically generate log messages stating when the method of some service has been entered and left.
+
+```scala
+@derive(representableK, loggingMid)
```
patseev, line 22: maybe providing necessary imports would be useful here

docs/logging/main-entities.md
```diff
@@ -0,0 +1,91 @@
...
+```json
+{
+  "@timestamp":"2021-08-20T17:13:39.787Z",
+  "loggerName":"some-logger",
+  "threadName":"ioapp-compute-0",
+  "level":"INFO",
+  "message":"This is Payment: Payment{id=3234,cardNumber=...,name=ClientData{name=foo,surname=bar}}",
+  "id":3234,
+  "cardNumber":"...",
+  "name":{
```
patseev, line 52: `@unembed` in case class definition means that here `name` and `surname` should be one layer higher, right?

patseev (commented):
Would be cool to hint that everything works out of box with ZIO as well

FunFunFine:
> Would be cool to hint that everything works out of box with ZIO as well

Sadly, I don't have that much experience with ZIO to write myself a full example for that. I will add a hint to check out the ZLogs module anyway.

docs/logging/home-page.md
FunFunFine: There are some crumbs of information yet to be transferred to the new documentation. Things don't happen here that fast to remove obsolete one completely :(

PR #735, opened 2021-08-07
Fix tapError signature in EnvBio

Paulymorph:


Paulymorph:
В `Suite` в `handle` есть еще такой странный тест
```scala
  it should "return new error if effectful computation fails" in {
    EnvBio
      .raiseError("Err1")
      .tapHandle(_ => EnvBio.raiseError("Err2"))
      .run(())
      .runSyncUnsafe(Duration.Inf) shouldBe Left("Err1")
  }
```
В названии одно, при этом проверяется на первую ошибку. Как правильно?

modules/env/src/main/scala/tofu/env/bio/EnvBio.scala
```diff
@@ -72,7 +72,7 @@ abstract class EnvBio[-R, +E, +A] {
   def mapError[E1](f: E => E1): EnvBio[R, E1, A] =
     onErrorHandleWith(e => EnvBio.raiseError(f(e)))
 
-  def tapError[R1 <: R, E1 >: E, A1 >: A](f: E => EnvBio[R1, E1, A1]): EnvBio[R1, E1, A1] =
+  def tapError[R1 <: R, E1 >: E](f: E => EnvBio[R1, E1, _]): EnvBio[R1, E1, A] =
```
road21, line 75: I guess `EnvBio[R1, E1, Any]` would be better here

Paulymorph: I have `_` here as a means to explicitly say something like `I don't care about particular type of this type parameter`. I am not quite sure if that makes sense here to have `Any`. I guess, that for compilation it should not matter a lot. So I am not sure which should be better.
 
I would consider an alternative where the method had the following signature in case it saved binary compatibility.
```scala
def tapError[R1 <: R, E1 >: E, B](f: E => EnvBio[R1, E1, B]): EnvBio[R1, E1, A]
```
But I am not sure how could I check that.

Paulymorph: Ok, here we are going to change the return type of the method. As I understood the [spec ](https://docs.oracle.com/javase/specs/jls/se7/html/jls-13.html) 13.4.15 says that we are getting a new method anyway as we are changing the return type and that is not binary compatible.

Odomontois: Changing type argument of paremetric result doesn't break bincompat. Them all are just `EnvBio` for the JVM

Odomontois:
> В `Suite` в `handle` есть еще такой странный тест
> 
> ```scala
>   it should "return new error if effectful computation fails" in {
>     EnvBio
>       .raiseError("Err1")
>       .tapHandle(_ => EnvBio.raiseError("Err2"))
>       .run(())
>       .runSyncUnsafe(Duration.Inf) shouldBe Left("Err1")
>   }
> ```
> 
> В названии одно, при этом проверяется на первую ошибку. Как правильно?
Нужно изменить описание теста. 
На мой взгляд, это корректная работа `tapHandle`

Odomontois (approved):

modules/env/src/main/scala/tofu/env/bio/EnvBio.scala
Paulymorph: @Odomontois what about `Any` vs `_`?

Odomontois: > @Odomontois what about `Any` vs `_`?

For a covariant TC F[Any] is nearly the same as F[_], but F[_] can be target of bizarre typecheck bugs sometime, so I prefer Any

Paulymorph: Fixed

FunFunFine (approved):

PR #740, opened 2021-08-18
Kernelize agent

FunFunFine:
This PR moves Agent & SerialAgent to the kernel module, and defines separate MakeRef and MakeSemaphore for CE2 & CE3.
 
Also, I deleted `Agents[F]` in favour of  `Agent.Make[F]` for two reasons:
- to keep close to CE3 `Ref.Make[F]` and such
- to avoid package object clashes.


Odomontois:
Looks great.
But needs  summon tests for both ce2\ ce3

Odomontois (approved):

PR #762, opened 2021-09-15
Remove deprecations from docs (hascontext)

Ssstlis:


docs/hascontext.md
```diff
@@ -1,16 +1,17 @@
 ---
-id: hascontext
-title: HasContext
+id: withcontext
```
FunFunFine, line 2: Filename must be the same as id, otherwise docusaurus won't show the page.

FunFunFine (approved):

docs/hascontext.md
Ssstlis: heck

Ssstlis: fixed

catostrophe (approved):

PR #766, opened 2021-09-17
ELKLayout read static json from customFields tag

Phill101:


Odomontois:
That's great, but a small docstring above `def setCustomFields` about usage would be cool

Odomontois (approved):

Odomontois (approved):

PR #768, opened 2021-09-17
fix PlusLoggable, add SingletonLoggable

### LINKED ISSUE #759: Poor loggable composition
An expected behaviour of the `Logging.named` is to embed existing fields into a single external field.
Meanwhile `Loggable` composition via `+`, followed by `named`, produces unexpected results:

```scala
import tofu.logging._
import cats.syntax.semigroup._

def inspect(loggable: Loggable[String]) = println(TethysBuilder.apply[String]("")(loggable))

val fooLog = Loggable.stringValue.named("foo")
val barLog = Loggable.stringValue.named("bar")

val log1 = fooLog + barLog
val log2 = new DictLoggable[String] with HideLoggable[String] {
  def fields[I, V, R, S](a: String, i: I)(implicit r: LogRenderer[I, V, R, S]): R =
    fooLog.fields(a, i) |+| barLog.fields(a, i)
}
inspect(log1) // good: {"foo":"","bar":""}
inspect(log2) // good: {"foo":"","bar":""}


inspect(log1.named("buz")) // unexpected: {"buz":"","buz":""}
inspect(log2.named("buz")) // good:       {"buz":{"foo":"","bar":""}}
```
[scastie](https://scastie.scala-lang.org/nA2mpEWJRZa1GUMBPcFWgQ)

Odomontois:
resolve #759

### DIFF
--- a/modules/logging/structured/src/main/scala/tofu/logging/Loggable.scala
+++ b/modules/logging/structured/src/main/scala/tofu/logging/Loggable.scala
@@ -35,6 +35,11 @@ trait Loggable[A] extends Loggable.Base[A] {
   /** whenever fields  are called it would be a single field named `name` and corresponding value */
   def named(name: String): Loggable[A] = new NamedLoggable[A](name, this)
 
+  /** something that works as [[named]] on the toplevel but ensures, that field is always represented as a singleton
+    * dict inside other value
+    */
+  def singleton(name: String): Loggable[A] = new SingletonLoggable[A](name, this)
+
   def showInstance: Show[A] = logShow
 
   def narrow[B <: A]: Loggable[B] = this.asInstanceOf[Loggable[B]]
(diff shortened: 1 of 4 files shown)


Odomontois:
#759 found several issues in the composition via `+`

1. `PlusLoggable` redefined `putField` in a way incompatible with `putValue`. It tried basically to append both fields with the same name which is nonsensical
2. Current `named` modification otherwise behaves correctly. It is primarily a modification for a value to be contextually logged with a preferred name, it must keep the same behavior when it is seen as someone's field. Still, it could be pleasant for the user to see that it intended to transform the Loggable to a singleton dict. So we add the `singleton`
3. Again current implementation of `putValue` for the `PlusLoggable` attempts to put some value in the current position for the first loggable and only tries the second loggable if the first fails. It makes sense since we can't combine arbitrary values. But when we talk about DictLoggable it guarantees the success of `putValue`, with at least empty dict. So it could be great for it to push additional fields to the dict when the second Loggable wants it. So instead we delegate the plus's `putValue' method to a special method `combinedValue` in the Loggable. It defaults to the previous behaviour, i.e. coalesce, but will try to extend DictLoggable from the second Loggable when it's used inside the `PlusLoggable`

danslapman (approved):
LGTM

PR #782, opened 2021-09-21
Upgrade ZIO Logging

vagroz:
This PR brings changes of tofu-zio-logging according to the new Module Pattern 2.0. Plain logging with encapsulated FiberRef context will be especially useful. 
I've written the documentation how to use it. Pls check my grammar.

Also fix logging documentation:
1. move tofu.logging.recipes to Receipes category
2. make page Loggable typeclass as copy-paste from the old docs


Odomontois (approved):

docs/tofu.logging.recipes.zio.md
```diff
@@ -0,0 +1,168 @@
...
+    } yield ()
+  }.injectCustom(
+    BarService.live,
+    logsLayer,
+    contextLayer
+  ).exitCode
+}
+```
+Run the program and look at the output:
+```json lines
```
FunFunFine, line 162: Perhaps it's better to write this example using ConsoleContextLayout for readability. Even on 23'' screen I have to scroll it right now to see the main stuff.

docs/tofu.logging.recipes.zio.md
```diff
@@ -0,0 +1,168 @@
...
+Let's consider how to log the context along with actual log message. If you have an instance of `Loggable` for your context, 
+you can have it logged automagically. For example:
+```scala
+import derevo.derive
+import tofu.logging.derivation.loggable
+
+@derive(loggable)
+final case class Context(requestId: Int)
+```
+Here we use [Tofu Derevo](https://github.com/tofu-tf/derevo) for automatic derivation `Loggable` instance.
```
FunFunFine, line 131: It would be nice to see the link to Loggable page here.

docs/tofu.logging.recipes.zio.md
```diff
@@ -0,0 +1,168 @@
...
+```scala
+import tofu.logging.zlogs._
+```
+This contains some useful stuff:
+
+##### `tofu.logging.Logging` type aliases
+These services do logs.
+- `ULogging` — is a type alias for `Logging[UIO]`. Logging methods have no environment like 
+`def info(message: String, values: LoggedValue*): UIO[Unit]`
+- `ZLogging[R]` — is a type alias for `Logging[URIO[R, *]]`. Logging methods require a ZIO environment `R`:
```
FunFunFine, line 22: It's hard to tell why we would need two different `Logging`s. I guess something like "ZLogging is for context logging, ULogging is for everything else" would be enough.

docs/tofu.logging.recipes.zio.md
```diff
@@ -0,0 +1,168 @@
...
+```scala
+import derevo.derive
+import tofu.logging.derivation.loggable
+
+@derive(loggable)
+final case class Context(requestId: Int)
+```
+Here we use [Tofu Derevo](https://github.com/tofu-tf/derevo) for automatic derivation `Loggable` instance.
+
+The main idea of this approach is to store your in ZIO [FiberRef](https://zio.dev/docs/datatypes/fiber/fiberref).
```
FunFunFine, line 133: What if someone would want to store the Context not in FiberRef, but in R itself (which is blasphemy of course)? Should we have a note about this approach here?

modules/zio/logging/src/main/scala/tofu/logging/impl/ZUniversalContextLogging.scala
```diff
@@ -0,0 +1,22 @@
+package tofu.logging.impl
+
+import org.slf4j.{LoggerFactory, Marker}
+import tofu.logging.{LoggedValue, Logging}
+import zio._
+
+class ZUniversalContextLogging[R](name: String, fctx: (LoggedValue => Unit) => URIO[R, Unit])
```
FunFunFine, line 7: > fctx: (LoggedValue => Unit) => URIO[R, Unit]

We definitely need a scaladoc with a note on unsafety of this argument

modules/zio/logging/src/test/scala/tofu/logging/zlogs/ZLogsSuite.scala
```diff
@@ -37,6 +39,34 @@ class ZLogsSuite extends AnyFlatSpec with Matchers {
       LogTree(ctx)
     } should ===(List.fill(2)(expected))
   }
+
+  "ZLogs plain with context" should "log the context" in {
+    type CtxService = FiberRef[(FooService, BarService)]
```
FunFunFine, line 44: Is it possible to rework this spec? It's kinda confusing when I see context which is named as some service or something.

docs/tofu.logging.recipes.zio.md
vagroz: What do you mean by "context logging"? `plainLoggingWithContext` is context logging, but encapsulates the context inside, so this is `ULogging`

docs/tofu.logging.recipes.zio.md
vagroz: Yes, we should. But I've never use the R as a context. It may be useful with TF and WithContext monad, but I'm not good in TF, so I'm not able to write a good example with that approach.

Now I can note that this is one of possible ways

docs/tofu.logging.recipes.zio.md
FunFunFine: Ah, I see now.

modules/zio/logging/src/test/scala/tofu/logging/zlogs/ZLogsSuite.scala
vagroz: > context which is named as some service
Look, actually `FiberRef[(FooService, BarService)]` is a service providing the context `(FooService, BarService)`, right?
From ZIO docs: 
> Has[A] represents a dependency on a service of type A
And `val logLayer: URLayer[Has[CtxService], Has[ZLogging.Make]]` below depends on this service
In production the service would be more complex like
```
trait CtxService {
  def getContext: UIO[(GlobalSate, FiberState)]

  def setReadiness(v: Boolean): UIO[Unit]

 //... etc
}
```

docs/tofu.logging.recipes.zio.md
vagroz: I think it's ok just remove extra fields

modules/zio/logging/src/main/scala/tofu/logging/impl/ZUniversalContextLogging.scala
Odomontois: I feel we can simplify that down to 
`ctxLog: URIO[R, LoggedValue]`
and use direct flatMap in the logger

vagroz: Oh, I've got it. I'll fix

vagroz: Done

Odomontois (approved):

PR #787, opened 2021-09-23
Fix unused warnings in logging

Odomontois:
Some warnings introduced in #782, somehow missed


modules/zio/logging/src/test/scala/tofu/logging/zlogs/ZLogsSuite.scala
```diff
@@ -53,6 +54,7 @@ class ZLogsSuite extends AnyFlatSpec with Matchers {
     val logLayer: URLayer[Has[CtxService], Has[ZLogging.Make]] =
       ZLogging.Make.layerPlainWithContext(_.get)
 
+    @nowarn("msg=local val .+ is never used")
```
catostrophe, line 57: Replace with `@nowarn("cat=...")`

Odomontois: What is the cat?

PR #798, opened 2021-10-03
Implicit conversion from Bind to cats.Monad

gzoom13:
From user perspective, when I require Bind instance in my code,
I want to have implicit cats.Monad instance for later usages. Now to do so
I have to write something like implicit def monad[E] = bind.monad[E]
in the beginning of the method, since most of the library methods accept
typeclasses as implicit parameters.

Assuming that Bind is right-biased, it makes perfect sense to write such
implicit conversion and use it in source file imports section.

### DIFF
--- a/modules/higherKindCore/src/test/scala/tofu/control/CatsInstancesFromBindSuite.scala
+++ b/modules/higherKindCore/src/test/scala/tofu/control/CatsInstancesFromBindSuite.scala
@@ -1,6 +1,6 @@
 package tofu.control
 
-import cats.FlatMap
+import cats.{Applicative, FlatMap}
 import tofu.compat.unused
 import tofu.instances.bind._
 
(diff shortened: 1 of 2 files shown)


modules/higherKindCore/src/test/scala/tofu/control/CatsInstancesFromBindSuite.scala
```diff
@@ -0,0 +1,16 @@
...
+import tofu.compat.unused
+import tofu.instances.bind._
+
+class CatsInstancesFromBindSuite {
+  def summonBindInstances[F[+_, +_]](implicit bind: Bind[F]): Unit = {
+    requireApplicative[F[Throwable, *]]
+    requireFlatMap[F[Nothing, *]]
+  }
+
+  def requireApplicative[F[_]](implicit @unused applicative: FlatMap[F]): Unit = ()
```
Ssstlis, line 13: Actually you summon FlatMap, not an Applicative

modules/higherKindCore/src/test/scala/tofu/control/CatsInstancesFromBindSuite.scala
```diff
@@ -0,0 +1,16 @@
...
+
+  def requireFlatMap[F[_]](implicit @unused applicative: FlatMap[F]): Unit = ()
```
Ssstlis, line 15: Wrong naming

modules/higherKindCore/src/test/scala/tofu/control/CatsInstancesFromBindSuite.scala
gzoom13: Thank you, fixed

modules/higherKindCore/src/test/scala/tofu/control/CatsInstancesFromBindSuite.scala
gzoom13: Fixed

Odomontois (approved):

PR #899, opened 2022-02-10
Add a tofu logging provider for logstash logback encoder

vilunov:


FunFunFine (approved):
We do need documentation for that though.

Odomontois (approved):

PR #926, opened 2022-05-01
BoundedParallel typeclass

vilunov:
Wander typeclass for bounded parallel traverse with implementations for CE2 and CE3


modules/kernel/src/main/scala/tofu/Wander.scala
```diff
@@ -0,0 +1,18 @@
+package tofu
+
+import cats.Traverse
+import tofu.internal.carriers.{WanderCarrierCE2, WanderCarrierCE3}
+
+trait Wander[F[_]] {
```
FunFunFine, line 6: Monix has deprecated the name, and so should we. 
I believe it will confuse newcomers more, as Monix itself as a library is not that popular anymore.

vilunov: Like we discussed, it's not essential to name it Wander, but I need to have specific proposals for a new name, otherwise I'll keep this one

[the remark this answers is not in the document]
vilunov: We decided not to do this, maybe later when we'll come to implementing something similar to Parallel

Odomontois (approved):

PR #1001, opened 2022-08-29
Add discard method as a safer alternative to void

FunFunFine:
Check out the scaladoc for motivation.


vilunov:
one small unit test and then it will LGTM

Odomontois (approved):

PR #1038, opened 2022-11-15
ZIO-2 logging (Part 1)

vagroz:
This PR copy-pastes ZIO-1  implementation with small improvements. 
The docs and ZIO-Native implementation will be brought with another PR.


examples-zio2/src/main/scala/tofu/example/logging/ZMakeSimpleExample.scala
```diff
@@ -0,0 +1,26 @@
...
+import zio._
+
+object ZMakeSimpleExample extends ZIOAppDefault {
+
+  val tasks = (1 to 3)
+    .map(i =>
+      for {
+        id <- Random.nextIntBetween(1000, 4000)
+        _  <- ZIO.serviceWithZIO[SimpleContext](_.set(id.toHexString))
+        _  <- ZIO.serviceWithZIO[FooService](_.foo(i))
```
FunFunFine, line 14: Most of the readers will look at this example only from github web UI, so it makes sense to add `expecation` comments here, like 
```suggestion
        _  <- ZIO.serviceWithZIO[FooService](_.foo(i)) //logs something 
```

examples-zio2/src/main/scala/tofu/example/logging/impl/SimpleContext.scala
```diff
@@ -0,0 +1,23 @@
...
+
+import tofu.logging.{Loggable, LoggedValue}
+import tofu.logging.zlogs.ContextProvider
+import zio.{FiberRef, UIO, ULayer, ZLayer}
+
+class SimpleContext(ref: FiberRef[String]) extends ContextProvider {
+  private implicit val requestIdLoggable: Loggable[String] =
+    Loggable.stringValue.named("requestId")
+
+  override def getCtx: UIO[LoggedValue] = ref.get.map(x => x)
```
FunFunFine, line 11: Is there any other way to get implement this method other than that? It feels like any implementation of `ContextProvider` will be doing `something.something.map(x => /*implicit conversion*/ x)`. Perhaps it's better to make it `ContextProvider[A: Loggable]`?

examples-zio2/src/main/scala/tofu/example/logging/impl/SimpleContext.scala
```diff
@@ -0,0 +1,23 @@
...
+
+  def set(requestId: String): UIO[Unit] =
+    ref.set(requestId)
+}
+
+object SimpleContext {
+  val layer: ULayer[SimpleContext] = ZLayer.scoped(
+    FiberRef
+      .make[String]("")
```
FunFunFine, line 20: I understand that this just an example, but it's still kinda not cool to have it like that IMO. At least put there something like `uninitialised Request Id`

modules/zio2/logging/src/main/scala/tofu/logging/impl/ZContextLogging.scala
```diff
@@ -0,0 +1,34 @@
+package tofu.logging.impl
+
+import org.slf4j.{Logger, Marker}
+import tofu.logging.zlogs.ZLogging
+import tofu.logging.{Loggable, LoggedValue, Logging}
+import zio._
+
+class ZContextLogging[R, Ctx: Loggable](logger: => Logger, ctxLog: URIO[R, Ctx]) extends ZLogging[R] {
```
FunFunFine, line 8: >logger: => Logger

why by-name?

vagroz: Discussed with @Odomontois this `by-name` is OK

examples-zio2/src/main/scala/tofu/example/logging/impl/SimpleContext.scala
vagroz: `ZLogs.layerPlainWithContext` is pretty now: we can provide them any implementation of `ContextProvider`.
If we make this trait as `ContextProvider[A: Loggable]`, the layer will also require  type-parameter which can be huge enough. For example, next PR will bring `ContextProvider` with data `A = Map[ZLogAnnotation[_], Any]`. 

As a user, I don't want write THIS type `A`, ideally I should not know anything about it. I just write two things
```
zioProgram.provide(
  ZLogs.layerPlainWithContext,  // dear tofu, give me a logger factory using ContextProvider I point next to
  TofuDefaultContext.layer         // I'd like to log this context, don't carry how this collected
)
```

vagroz: > will be doing something.something.map(x => /*implicit conversion*/ x)

suggest `abstract class ValueContextProvider` which do this itself

FunFunFine: > suggest abstract class ValueContextProvider which do this itself

yep, that'll work

FunFunFine (approved):

PR #1074, opened 2022-12-16
Zio2 logging: tests + documentation

vagroz:


modules/interop/zio2/logging/src/test/scala-2/tofu/logging/zlogs/ZLogsSpec.scala
```diff
@@ -0,0 +1,114 @@
...
+import tofu.logging.impl.ContextMarker
+import tofu.logging.zlogs.TestStuff._
+import tofu.logging.zlogs.TofuDefaultContextSpec.{testCount, testStatus, testUser}
+import tofu.logging.{LogTree, LoggedValue}
+import zio._
+import zio.test._
+
+object ZLogsSpec extends ZIOSpecDefault {
+
+  val loggerName: String = this.getClass.getName.replace("$", "")
```
FunFunFine, line 14: `case object ZLogsSpec` & `this.productPrefix`

docs/tofu.logging.recipes.zio2.md
```diff
@@ -0,0 +1,213 @@
...
+### ContextProvider
+
+```scala
+trait ContextProvider {
+  def getCtx: UIO[LoggedValue]
+}
+```
+
+Is required by `layerPlainWithContext` factory, every logger will log the provided `LoggedValue`.
+To implement your own instance it can be convenient to use `ValuedContextProvider`:
```
FunFunFine, line 141:
```suggestion
It can be convenient to use `ValuedContextProvider` to implement your own instance:
```

docs/tofu.logging.recipes.zio2.md
```diff
@@ -0,0 +1,213 @@
...
+
+### ContextProvider
+
+```scala
+trait ContextProvider {
+  def getCtx: UIO[LoggedValue]
+}
+```
+
+Is required by `layerPlainWithContext` factory, every logger will log the provided `LoggedValue`.
```
FunFunFine, line 140: I think we need a bit more motivation/description here. At first glance it seems like a static context provider or something.

FunFunFine:
It'd be nice to have the same "recipe" for ZLogs as for the [other logging approaches ](https://github.com/tofu-tf/tofu/tree/master/examples/ce2/src/main/scala-2/tofu/example/logging). It will both illustrate the intended way and will allow people to look at this as a whole thing with wiring and stuff. 
Basically it should be the copy of that cargo API, but without TF and in idiomatic ZIO way.

 It'd be fine if it comes in another PR though

FunFunFine (approved):

modules/interop/zio2/logging/src/test/scala-2/tofu/logging/zlogs/ZLogsSpec.scala
vagroz: this should be with package prefix

PR #1154, opened 2023-07-07
zio-logging improvements

zaxxel:
- added slf4j initialization call, because in some cases of program failure it may happen that the program exits before the log message is written.
- scoped log annotations added for convenience

### DIFF
--- a/docs/tofu.logging.recipes.zio2.md
+++ b/docs/tofu.logging.recipes.zio2.md
@@ -82,7 +82,7 @@ will produce:
 ```json
 {
   "level": "INFO",
-  "loggerName": "MyLogger",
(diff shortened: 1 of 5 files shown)


docs/tofu.logging.recipes.zio2.md
```diff
@@ -82,14 +82,34 @@ will produce:
 ```json
 {
   "level": "INFO",
-  "logger_name": "MyLogger",
+  "loggerName": "MyLogger",
```
vagroz, line 85: `logger_name` is default field of logback-logstash. This spelling is using in the other examples, return back please

Odomontois (approved):

PR #1188, opened 2023-10-19
fix: Masking for contramapped loggables

Mee-Tree:
Currently `ContramapLoggable` doesn't respect masking
These changes are meant to fix that

The previous behavior can be seen here:
https://scastie.scala-lang.org/Rw5jMFJ2Q7OukcdgVRXaUg


dos65 (approved):
Thanks for the fix!

Issue #1068, opened 2022-12-01
Scala 3 tracking Issue


Odomontois:
- [ ] Migration of higher-kind core
- [ ] Migration of kernel
- [ ] Migration of higher-kind macros
- [ ] Migration of ce2 support
- [ ] Migration of ce3 support
- [ ] Migration of ZIO support
- [ ] Migration of logging

Roman-Statsura:
I'll take on task for migration zio-logging on Scala 3, if no one minds

dos65:
Attaching the curent status of having Scala3 support for each module :
- [ ] concurrentCE2
- [ ] config
- [x] coreCE2
- [x] coreCE3
- [ ] doobie
- [ ] doobieCE3
- [ ] doobieLogging
- [ ] doobieLoggingCE3
- [ ] env
- [ ] fs2CE2Interop
- [x] fs2CE3Interop
- [x] higherKindCore
- [x] kernel
- [x] kernelCatsMtlInterop
- [x] logging
- [x] loggingDer
- [x] loggingEnumeratum
- [x] loggingLayout
- [x] loggingLog4Cats
- [ ] loggingLog4CatsLegacy
- [x] loggingLogstashLogback
- [x] loggingRefined
- [ ] loggingShapeless
- [x] loggingStr
- [ ] memo
- [ ] observable
- [x] streams
- [ ] zio1Core
- [ ] zio1Interop
- [ ] zio1Logging
- [ ] zio2Core
- [x] zio2Logging

PR #1134, opened 2023-05-31
Scala 3 build for higherKindCore module

DoSOfRedRiver:
Here some WIP on porting higherKindCore to Scala.

PR #1192, opened 2023-11-20
Scala3/kernel+core

Grryum:
Scala 3 is coming!

Here are modules already ported:
- kernel-higherKind (all except macros) - based on #1134 
- kernel (all RepresentableK instances derived by hand without hk macros)
- coreCE2
- coreCE3
- zio2Core


modules/kernel/src/main/scala-3/tofu/internal/Interop.scala
```diff
@@ -0,0 +1,89 @@
...
+            withExplicitArgs.appliedToArgs(summonedInst)
+          case _             =>
+            withExplicitArgs
+
+        // for some reason lambda type like this `FinallyCarrier2.Aux[F, Throwable, [x] =>> CE2Kernel.CEExit[java.lang.Throwable, x]]`
+        // has no subtyping relations with type `FinallyCarrier2.Aux[F, Throwable, [x] =>> Any]` and `.asExprOf[X]` throws java.lang.Exception: Expr cast exception
+
+        // For such cases old and classic `asInstanceOf` helps with extra casting.
+
+        val asInstanceOf = TypeRepr.of[AnyRef].typeSymbol.methodMember("asInstanceOf").head
```
Odomontois, line 75: That's very unpleasant. But I suppose we can cope with it for now

Odomontois:
Thank you very much for all of this colossal work.
The only thing that frightens me is the proposed implementation of [`Interop` macros](https://github.com/tofu-tf/tofu/pull/1192/files#diff-048ed709a654c7dd5bf69a659b0e83a13168c83b480ca5090b179a6ca5431e12).
It clearly was done to be compatible with scala 2 definitions.
When I've implemented this I've missed some features of scala 3 inlines a lot, meaning I had to pass code line to the macro via structural type.
So I've had in mind that when that would be migrating to scala 3, we will rewrite the delegation implementation and "delegating" to instances would have scala-specificб but because of all that huge work, I hope we can do that in the folliwing PR.
So I have no options but to accept this PR, it feels like it's now or never

Odomontois (approved):

modules/kernel/src/main/scala-3/tofu/internal/Interop.scala
Grryum: yup! @road21 tried to fix it and replaced with `widen`.
Hope now it is not so unpleasant :)

Grryum:
partially closes #1068

Odomontois:
What happened with doc generation?

PR #1234, opened 2024-04-08
feature/add fold* to EitherOps

DZRock:
Hello! I didn't find any impl for plain Either.fold so i suggest this

### DIFF
--- a/modules/core/ce2/src/test/scala/tofu/syntax/FEitherSuite.scala
+++ b/modules/core/ce2/src/test/scala/tofu/syntax/FEitherSuite.scala
@@ -406,6 +406,20 @@ class FEitherSuite extends AnyWordSpec with Matchers {
     }
   }
 
+  "EitherFOps#foldIn" should {
+    "map left or right to C" in {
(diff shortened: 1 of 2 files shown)


modules/core/ce2/src/test/scala/tofu/syntax/FEitherSuite.scala
```diff
@@ -406,6 +406,20 @@ class FEitherSuite extends AnyWordSpec with Matchers {
...
+    "map left or right to C" in {
+      defaultRight.foldIn(_.length, c => c * c) mustBe Some(16)
+      defaultLeft.foldIn(c => c * c, _.length) mustBe Some(16)
+    }
+  }
+
+  "EitherFOps#foldF" should {
+    "map left or right to F[C]" in {
+      defaultRight.foldF(s => Some(s.length), c => Some(c * c)) mustBe Some(16)
+      defaultLeft.foldF(c => Some(c * c), s => Some(s.length)) mustBe Some(16)
```
FunFunFine, line 419: let's add a couple of test cases with `c => None`

FunFunFine (approved):

PR #1238, opened 2024-04-12
fixed zio.Tag's usage in ZIO's WithRun implicit

### LINKED ISSUE #1236: ZIO WithRun#runContext fails with an error
The following code fails with `java.lang.Error: Defect in zio.ZEnvironment: Could not find ErrorExample::Context inside ZEnvironment(Any -> Context(42))`
```
import tofu.WithRun
import tofu.zioInstances.implicits.zioTofuWithRunImplicit
import zio._

object ZIOWithRunError extends ZIOAppDefault {
  case class Context(x: Int)

  val app: RIO[Context, Unit] = ZIO.service[Context].flatMap(Console.printLine(_))

  override def run: ZIO[Any with ZIOAppArgs with Scope, Any, Any] =
    WithRun[RIO[Context, *], Task, Context].runContext(app)(Context(42))
}
```
Based on the error message, it seems like `Context`'s tag is not used to build `ZEnvironment` when using `WithRun` instance, resulting in ZIO being unable to find `Context` instance in the environment.

skelantros:
- An implicit instance of `Tag` is now used in ZIO's `WithRun` instance to initialize `ZEnvironment` properly.

fixes #1236

### DIFF
--- a/modules/interop/zio2/core/src/main/scala/tofu/zioInstances/ZioInstances.scala
+++ b/modules/interop/zio2/core/src/main/scala/tofu/zioInstances/ZioInstances.scala
@@ -46,9 +46,7 @@ private[zioInstances] class ZioInstances {
   private[this] val zioTofuInstanceAny: ZioTofuInstance[Any, Any] = new ZioTofuInstance
   final def zioTofuInstance[R, E]: ZioTofuInstance[R, E]          = zioTofuInstanceAny.asInstanceOf[ZioTofuInstance[R, E]]
 
-  private[this] val zioTofuWithRunInstanceAny                          = new ZioTofuWithRunInstance[Any, Any]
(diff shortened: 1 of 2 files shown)


Grryum:
nice fix!

dos65:
@skelantros Thx for the fix!  

It would be nice if there was a test on that issue to be sure that we won't break it. 
Don't you mind to add that in your PR?

skelantros:
> @skelantros Thx for the fix!
> 
> It would be nice if there was a test on that issue to be sure that we won't break it. Don't you mind to add that in your PR?

Hi there! Ok, gonna do that on the weekend.

dos65:
@skelantros I'm closing your PR. Your changes were included in #1383  by @Grryum

PR #1242, opened 2024-04-16
feature: Add clockForKleisli and sleepForKleisli

geny200:
Hi, I suggest to add same functionality from cats for effects, so I opened a small pull request to demonstrate what I propose. If it's ok, then in the new PR I'll do it for all transformers and all effects.

Here is an example with a demonstration of the problem: In the first case, with cats everything is fine; in the second case, the ide says that everything is fine, but we get a compilation error.

```scala
  import cats.effect.{Clock => CatsClock}

  // Ok
  def catsClock[F[_]: Monad: CatsClock]: F[FiniteDuration] = {
    val clock = CatsClock[ReaderT[F, String, *]]
    clock.realTime.run("str")
  }

  import tofu.time.{Clock => TofuClock}
  
  // Error `could not find implicit value for parameter e: tofu.time.Clock[[â•¬â”‚$1$]cats.data.Kleisli[F,String,â•¬â”‚$1$]]`
  def tofuClock[F[_]: Monad: TofuClock]: F[Long] = {
    val clock = TofuClock[ReaderT[F, String, *]]
    clock.realTime(SECONDS).run("str")
  }
```

### DIFF
--- a/modules/core/ce2/src/test/scala/tofu/time/TimeSuite.scala
+++ b/modules/core/ce2/src/test/scala/tofu/time/TimeSuite.scala
@@ -34,7 +34,7 @@ class TimeSuite {
     Timeout[ReaderT[IO, Unit, _]]
   }
 
-  def readerF[F[_]: Clock : Sleep] = {
+  def readerF[F[_]: Clock: Sleep] = {
     Clock[ReaderT[F, Unit, _]]
     Sleep[ReaderT[F, Unit, _]]
   }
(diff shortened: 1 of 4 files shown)


dos65 (approved):

dos65:
Thx for contribution! 

> If it's ok, then in the new PR I'll do it for all transformers and all effects.

Yeah. I think it's a good idea to provide instances for transfomers as CE does.

PR #1244, opened 2024-04-18
feature: Add ability to define custom masker function for @masked annotation

NarekDW:
I'd like to propose an important feature - ability to define a custom masker function for fields.
Let's say I'd like to mask different fields - differently. Like this:
```scala
@derive(loggable)
case class MaskedCustom(
      @masked sensitiveField: String, // this should be replaced with "*"
      @masked firstName: Option[String], // this should be replaced with "J***"
      @masked age: Int // this should be replaced with "**"
  )
```

With that purpose - I'm proposing to define a new `MaskMode`
```scala
  /** Allows to define a custom masker function */
  case class Custom(f: String => String) extends MaskMode
```
with high order custom  function


dos65 (approved):
LGTM!

PR #1245, opened 2024-04-18
Add transformers support for effects

geny200:
In this PR, I am extending the functionality for transformers as suggested in PR https://github.com/tofu-tf/tofu/pull/1242

Transformers support for:
* Clock
* Sleep
* TimeZone
* Delay
* GenRandom
* GenUUID

I plan to add support for transformers for Effects such as Timeout and Fire, but in a new PR, because there will be different logic from the current implementations


modules/kernel/src/main/scala-3/tofu/internal/instances/ClockInstance.scala
```diff
@@ -1,14 +1,26 @@
...
+import cats.~>
+import tofu.higherKind.{RepK, RepresentableK}
 import tofu.internal.carriers.ClockCE3Carrier
 import tofu.internal.carriers.ClockCE2Carrier
 import tofu.time.Clock
 
+import java.util.concurrent.TimeUnit
 import scala.compiletime.summonFrom
 
-private[tofu] trait ClockInstance:
```
dos65, line 10: all `*Instance` files uses optinal braces syntax. Could you return it back?

geny200: i've removed the braces

dos65 (approved):

PR #1262, opened 2024-05-01
Add support Scala 3 for modules streams and fs2CE3Interop

Roman-Statsura:
Add support Scala 3 for modules `streams` and `fs2CE3Interop`


dos65 (approved):
LGTM!

PR #1276, opened 2024-05-08
Update doobie dependency with new effectfull LogHandler[F[_]] and remove tofu embed log handler.

Grryum:
Since 1.0.0-RC3 doobie introduced completely reworked approach of using LogHandler. It became effectful & Transactor-level.
This pull request updates the Doobie dependency (for CE3) to accommodate this new effectful `LogHandler[F[_]]` and removes the Tofu embed log handler.
Previously, Doobie supported SQL statement logging by passing an implicit LogHandler to each run SQL method. After update, the LogHandler has been moved to the Transactor level, where it is defined once for the each transactor.
So there is no need for `EmbeddableLogHandler` and `LogHandlerF`.


dos65 (approved):
LGTM

PR #1277, opened 2024-05-08
Add support Scala 3 for module observable

Roman-Statsura:
Added support Scala 3 for module `observable` anded scala2And3Versions to jvmPlatform of `zio2Core` module


dos65 (approved):
Thx!

PR #1283, opened 2024-05-15
Fix bug with zioTofuBiInstance, add tests for zio2-core module, add logging mid test with bug related function for zio2 logging.

Grryum:


Grryum:
includes #1238

dos65 (approved):
LGTM

PR #1330, opened 2024-09-06
Add Permit type like Semaphore

geny200:
Hi, I noticed a disadvantage in tofu that there is no semaphore primitive, but there are concurrent primitives for atomic variables. To avoid creating semaphore based on `Agen`/`QVar`/`Atom`, I suggest to add a new primitive `Permit`, which will use native implementation of the semaphore, depending on the effects.

### DIFF
--- a/modules/core/ce2/src/main/scala/tofu/interop/CE2Agents.scala
+++ b/modules/core/ce2/src/main/scala/tofu/interop/CE2Agents.scala
@@ -40,7 +40,7 @@ final case class SerialSemRef[F[_]: Monad, A](ref: Ref[F, A], sem: Semaphore[F])
 
 /** Default implementation of [[tofu.concurrent.Permit]] that use [[cats.effect.concurrent.Semaphore]]
   */
-final case class PermitSem[F[_]](sem: Semaphore[F]) extends Permit[F] {
+final case class PermitSemaphore[F[_]](sem: Semaphore[F]) extends Permit[F] {
   def withPermit[A](fa: F[A]): F[A] = sem.withPermit(fa)
 }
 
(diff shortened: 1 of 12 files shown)


modules/core/ce3/src/test/scala/tofu/concurrent/PermitSuite.scala
```diff
@@ -0,0 +1,44 @@
...
+import tofu.syntax.monadic.*
+
+class PermitSuite extends AnyFunSuite {
+  private implicit val iort: IORuntime = IORuntime.global
+
+  @unused
+  private def summonInstance[I[_]: Sync, F[_]: Async]: MakePermit[I, F] =
+    implicitly[MakePermit[I, F]]
+
+  test("check IO has Atom") {
```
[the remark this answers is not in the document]
geny200: > IO has

this means that logic can evaluate on cats IO without any problems

modules/core/ce2/src/main/scala/tofu/interop/CE2Agents.scala
```diff
@@ -38,6 +38,12 @@ final case class SerialSemRef[F[_]: Monad, A](ref: Ref[F, A], sem: Semaphore[F])
     modifyM(a => if (f.isDefinedAt(a)) f(a) else (default, a).pure[F])
 }
 
+/** Default implementation of [[tofu.concurrent.Permit]] that use [[cats.effect.concurrent.Semaphore]]
+  */
+final case class PermitSem[F[_]](sem: Semaphore[F]) extends Permit[F] {
```
FunFunFine, line 43: I suggest using the full name, we don't gain anything from typing Sem instead of Semaphore

geny200: Hi, thx, I renamed it to PermitSemaphore

dos65 (approved):

PR #1333, opened 2024-09-12
Introduce logging-derivation-annotations package.

FunFunFine:
This MR allows adding Loggable derivation annontations to clases that are sensetive to the dependencies.


build.sbt
```diff
@@ -160,7 +160,16 @@ lazy val loggingDer = projectMatrix
     name := "tofu-logging-derivation"
   )
   .jvmPlatform(scala2And3Versions)
-  .dependsOn(loggingStr)
+  .dependsOn(loggingStr, loggingDerivationAnnotations)
+
+lazy val loggingDerivationAnnotations = projectMatrix
```
dos65, line 165: You also need to add this porjetc to aggregated `logging` module here - https://github.com/tofu-tf/tofu/blob/f8d62e574b0a9b6f807525dd604ad4fbfeaeaba0/build.sbt#L262 as `(loggingDerivationAnnotations, true, true)`

Otherwise it won't be published

dos65 (approved):

PR #1409, opened 2025-07-06
Add singleton enum Loggable instance

### LINKED ISSUE #1369: [Bug] Scala3 enums produces empty json in structured logs
## Reproduce:
Scala 3.3.5

https://scastie.scala-lang.org/nHE1epnTTXu27fUiKMSEvA\

### Bug:
```
... "message":"AppConfig{kek=Shrek,fiona=haha}","kek":{},"fiona":"haha" ...
```

### Expected
```
... "message":"AppConfig{kek=Shrek,fiona=haha}","kek":"Shrek","fiona":"haha" ...
```

Equwece:
Fixes #1369 by adding separate singleton enum Loggable instance which ensures that enum has only singleton cases and falls with compilation error otherwise.

```scala
enum NonSingletonEnum {
  case A
  case B(foo: Int)
}

val i1 = SingletonEnumLoggable.derived[NonSingletonEnum] // compilation error

enum SingletonEnum{
  case A
  case B
}

val i2 = SingletonEnumLoggable.derived[SingletonEnum] // ok
```

### DIFF
--- a/modules/logging/structured/src/main/scala-3/tofu/logging/SingletonEnumLoggable.scala
+++ b/modules/logging/structured/src/main/scala-3/tofu/logging/SingletonEnumLoggable.scala
@@ -1,6 +1,8 @@
 package tofu.logging
 
+import scala.deriving.Mirror
+
 object SingletonEnumLoggable:
-  inline def derived[T <: scala.reflect.Enum]: Loggable[T] =
+  inline def derived[T: Mirror.SumOf]: Loggable[T] =
     SingletonEnumLoggableMacro.ensureSingletonEnum[T]
     Loggable[String].contramap(_.toString)
--- a/modules/logging/structured/src/main/scala-3/tofu/logging/SingletonEnumLoggable.scala
+++ b/modules/logging/structured/src/main/scala-3/tofu/logging/SingletonEnumLoggable.scala
@@ -1,8 +1,0 @@
-package tofu.logging
-
-import scala.deriving.Mirror
-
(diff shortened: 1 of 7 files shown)


dos65 (commented):
Looks good!

Could remove unused imports from DerivedLoggableScala3Suite.scala ?

modules/logging/structured/src/main/scala-3/tofu/logging/SingletonEnumLoggable.scala
```diff
@@ -0,0 +1,8 @@
+package tofu.logging
+
+import scala.deriving.Mirror
+
+object SingletonEnumLoggable:
```
dos65, line 5: I missed one thing.

Could you also add `trait SignletonEnumLoggable[A] extends Loggable[A]`?

This would allow to use derives syntax

```
enum Foo derives SingletonEnumLoggable:
  case A
```

dos65: hah, yep. current impl with `toString` doen't look solid

The proper way to get names is to use Mirror. Here is an example:
```scala
inline def nameOf[A](using m: scala.deriving.Mirror.Of[A]): String = scala.compiletime.constValue[m.MirroredLabel]

enum Sample:
   case X
   
nameOf[Sample] // "Sample"
nameOf[Sample.X] // "Sample.X"  
```

dos65 (approved):

PR #1419, opened 2025-07-18
Use early-semver for publish

geny200:
See https://www.scala-lang.org/blog/2021/02/16/preventing-version-conflicts-with-versionscheme.html


dos65 (approved):
Thanks!

PR #1520, opened 2026-04-01
Fix TethysBuilder producing `{}` for Foldable types

### LINKED ISSUE #1358: TethysBuilder doesn't produce json from Foldable
Example
```scala
def json[A: Loggable](a: A) = TethysBuilder(a)

@derive(loggable)
final case class X(field: String)

val result = json(List(X("value")))
```

Expected:
```scala
[{"field":"value"}]
```

Actual
```scala
{}
```

I think this issue occurs due to how `fldLoggable` is constructed, particularly in this part:
```scala
def fields[I, V, R, M](a: A, input: I)(implicit receiver: LogRenderer[I, V, R, M]): R = input.noop
```

I'm not sure how this could be fixed without introducing significant changes...

bio-aeon:
Resolves #1358.

`TethysBuilder(List(X("value")))` returned `{}` instead of `[{"field":"value"}]` because `LogBuilder.apply` always wrapped output in a JSON object via `make`, but `SubLoggable.fields`, used by `Foldable` instances, is a noop.

This PR adds `putWrappedValue` to `Loggable.Base`, overridden in `SubLoggable`. `TethysBuilder.apply` now uses `putWrappedValue`.

```scala
TethysBuilder(List(1, 2, 3))  // was: {}  now: [1,2,3]
TethysBuilder(Foo("a", Some(1)))  // unchanged: {"str":"a","num":1}
```

### DIFF
--- a/modules/logging/derivation/src/test/scala/tofu/logging/derivation/DerivedLoggableSuite.scala
+++ b/modules/logging/derivation/src/test/scala/tofu/logging/derivation/DerivedLoggableSuite.scala
@@ -101,6 +101,10 @@ class DerivedLoggableSuite extends AnyFlatSpec with Matchers {
       "MaskedCustom{sensitiveField=*,firstName=Some(J***),age=**}"
   }
 
+  "top-level list of case classes" should "produce json array" in {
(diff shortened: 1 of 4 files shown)


bio-aeon:
As discussed in the Telegram channel, this fix contradicts the current design ideas of the library, and therefore I'm closing it.

PR #1526, opened 2026-04-12
Add `TethysBuilder.renderValue` without forced object wrapping

### LINKED ISSUE #1358: TethysBuilder doesn't produce json from Foldable

bio-aeon:
Resolves #1358.

`TethysBuilder(List(X("value")))` returns `{}` instead of `[{"field":"value"}]` because `LogBuilder.apply` always wraps output in a JSON object via `make`, but `SubLoggable.fields`, used for non-dict types like lists or ints, is a noop.
This PR adds `TethysBuilder.renderValue` that renders any loggable value in its natural JSON form.

```scala
TethysBuilder.renderValue(List(1, 2, 3))  // [1,2,3]
TethysBuilder(List(1, 2, 3))              // {} (unchanged)

### DIFF
--- a/modules/logging/derivation/src/test/scala/tofu/logging/derivation/DerivedLoggableSuite.scala
+++ b/modules/logging/derivation/src/test/scala/tofu/logging/derivation/DerivedLoggableSuite.scala
@@ -101,6 +101,10 @@ class DerivedLoggableSuite extends AnyFlatSpec with Matchers {
       "MaskedCustom{sensitiveField=*,firstName=Some(J***),age=**}"
   }
 
+  "TethysBuilder.renderValue" should "produce json array for list of case classes" in {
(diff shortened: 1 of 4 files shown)


dos65 (approved):

dos65:
LGTM, thanks!

PR #1576, opened 2026-08-09
Migrate tofu-doobie-ce3 to doobie 1.0.0-RC13 (org.typelevel)

bio-aeon:
Doobie moved the project to the Typelevel organization with two breaking changes:

| | groupId | root package |
|---|---|---|
| ≤ RC12 | `org.tpolecat` | `doobie.*` |
| RC13 | `org.typelevel` | `org.typelevel.doobie.*` |

This PR rewrites `doobie.` → `org.typelevel.doobie.`  across the CE3 modules: `tofu-doobie-ce3`, `tofu-doobie-logging-ce3` and the CE3 example.
CE2 modules are untouched.

There is also existing #1550 PR, which changes the `groupId` on all doobie-related modules, including the CE2 pair. This is wrong approach, since `doobie % 0.13.4` (the last doobie version with Cats Effect 2) was never published under `org.typelevel`.


dos65 (approved):
Great! Thanks for  the contribution!

